PR #110, opened 2021-10-06
Added Test Level for Pterodactyl Hazard

Fr4k26:
This level had an unfortunate bugs in it. Pterodactyls gets stuck in walls and also get stopped by the Wind Current ability. Also I can't seem to add a Respawn to this level.


Swiftshadow:
Old enough that merging it up is more difficult than its worth. Closing.

PR #130, opened 2021-10-21
Bc2 541 UI rework

CaptainVidia:
Deleted some UI for settings and menu scenes. Put them together for easier access.


Swiftshadow:
There are some material import problemss, but otherwise looks good!
Don't tie scene loading to a specific scene number, use the names instead. Scene numbers can change often, scene names won't
When moving things between projects, just copy the parts needed (just the scenes/scripts), not the entire project
    I copied the scenes/scripts/materials to the main project and nuked the imported one

PR #134, opened 2021-10-23
Bc2 572 change how wall scales

tcaron1024:
Changed how the wall scales to use dragging the mouse away from the click point rather than scrolling the mouse wheel. Fixed a small bug in vector placement script that caused the wall placement to use the slope placement on non sloped surfaces.  The new wall scaling doesn't work well on slopes - this is a larger issue with how vector placement on slopes work in general. I created another issue on Jira to change how vector placement on slopes works - going to be working on that for Tuesday.

Changed scripts:
WallScaling.cs 
VectorPlacement.cs


Swiftshadow:
Looks good!
What was the bug on slopes? I did not notice anything different

PR #172, opened 2021-11-02
Bc2 481 general upgrade system

QDonnelly-254:
Made the code here compliant with code regulations for the project


Swiftshadow:
UpgradeManager.cs: XML comment on lines 12-17 should be above the class, not namespace
UpgradeManager.cs line 235: should be Pause()
Other than those, looks good!
Note: updated your branch with main and resolved the merge conflicts that arose

PR #177, opened 2021-11-04
Bc2 582 meteor showing trajectory

tcaron1024:
Reworked how the meteor applies force and added in the trajectory line on the meteor decal to show where the ball will go. Fixed a bug with the standard placement decals not getting turned off when the ability is on cooldown.

Changed/new scripts:
StandardPlacement.cs (made it so decals get set inactive while the ability is on cooldown)
MeteorAbility.cs (changed how force gets added and added in a tag check)
MeteorUtilities.cs (new)
MeteorDecalTrajectory.cs (new)


Swiftshadow:
Trajectory line may be a bit long - thats configurable, so no issues
Love the documentation and commenting on MeteorUtilities.cs!

PR #179, opened 2021-11-04
Implemented thunderstorm hazard

brocEdson:
Implemented thundestorm hazard, with a test level and prefab created


Swiftshadow:
Its finally making it in!
Thunderstorm merged suprisingly nicely into the modern branch
Your OnTriggerStay does not feel quite efficient, but I cannot think of a better way at the moment

PR #180, opened 2021-11-04
Bc2 594 fix camera bugs

JesseCashmer:
Readded Keyboard Controls, adjustment in CameraControl Script.


Swiftshadow:
Zoom locking while in ballcam works!
    While not in ballcam, it does not
    Reversing the camera is still possible via zooming in outside of ballcam move, then entering it
The Invert checkboxes invert the keyboard controls too:
    Inverted X makes D take you left, and A right
    Inverted Y makes W take you backwards and S take you forwards
Did update main into your branch, and set the default boundaries

PR #182, opened 2021-11-04
BC2-645 hole complete anim and beam implemented

geekguy100:
Imported Spencer's hole beacon effect and win VFX. Added it to the prefab so it is accessible across all scenes.


Swiftshadow:
Looks nice!
You did have a missing tooltip (GolfHole.cs line 22), but I added it in to make sure this hits main

PR #183, opened 2021-11-04
Added a slippery slope physics material that has been added to the rib cage

QDonnelly-254:
Also made the ball red so we can see it better


Swiftshadow:
Had to change physics material to use the lowest friction, otherwise it looks good!

PR #184, opened 2021-11-04
Added close button for tutorial tooltips

brocEdson:
Also made the method for that a toggle to support pressing a button to toggle tutorial tooltips in the future. I would do it, but I don't know the specifics of the input system and don't want to mess with it without consulting with Pneuma.


Swiftshadow:
Close buttons do close stuff

PR #188, opened 2021-11-05
Bc2 651 vs adjustments

tcaron1024:
Added flag animation and object, reduced time delay on respawn, increased down force on ball while in air and falling.

New Scripts:
FlagBehavior.cs

changed:
PlayerRespawnManager.cs


Swiftshadow:
Looks good!

PR #189, opened 2021-11-05
BC2-651 level adjustments

tcaron1024:
fixed uneven ground near checkpoint 2


Swiftshadow:
Flat levels are good levels (unless otherwise slanted)

PR #191, opened 2021-11-05
Made Venus Flytrap only activate on touch

brocEdson:
What it says in the title.  Might need to fix up the physics bug for the flytrap because that looks like a major hindrance right now.


Swiftshadow:
Looks good!
Ball can still clip through the edges of the flytrap when it snaps, but that can be fixed for VS

PR #193, opened 2021-11-05
BC2-651 increasing max zoom amount

tcaron1024:
Increase max zoom amount on camera to fix bug


Swiftshadow:
Short and sweet, it works!

PR #195, opened 2021-11-06
BC2-651 getting rid of wind decal collider

tcaron1024:
Fix wind decal moving ball


Swiftshadow:
Wind decal no longer knocks the ball around

PR #197, opened 2021-11-06
Made the earthquake feel more powerful as a driver

QDonnelly-254:
The forwards force has been upgraded from 26 to 50. the wall now feels more viable as a bonus! charlis everywhere are thrilled!


Swiftshadow:
Earthquake be movin

PR #199, opened 2021-11-06
Bc2 659 programming the camera lock icon

brocEdson:
Added a camera lock icon that is activated when locked on and grays out when not locked on.


Swiftshadow:
No major conflicts with UI main!
It works!

PR #202, opened 2021-11-06
Made flytrap animation have physics, and made it lighter

brocEdson:
Made flytrap animation have physics, and made it lighter so it didn't launch the ball too far


Swiftshadow:
Better than before, still does launch the ball
    Make sure the animation changes get in
Accepting with the contition that the animation changes make it in before VS

PR #205, opened 2021-11-06
Bc2 661 camera flythrough level 1

tcaron1024:
Camera flythrough of level 1 complete. 

NOTE - LEVEL SELECT NEEDS TO LOAD LEVEL 1 FLYTHROUGH SCENE (IN SCENES/FLYTHROUGH FOLDER) 

new scripts:
FlythroughController.cs


Swiftshadow:
The flythrough looks very nice!
    It seems a bit stiff with the rotation, but thats a polish thing

PR #207, opened 2021-11-06
Bc2 651 vs adjustments

tcaron1024:
hole now sucks ball, made holes in tutorials 1/2 flush with ground, along with some changes you (doug) made

Scripts:
HoleSucc.cs


Swiftshadow:
Hungry Hungry Hole

PR #210, opened 2021-11-06
Bones and Sound Fixes

Mjsciarabba:


Swiftshadow:
Xylophone go brrrrrrrrr
Getting into the spine can be difficult, but when in there it does work!

PR #213, opened 2021-11-07
Bc2 667 camera shaking disasters

tcaron1024:
Implemented camera shake for meteor/earthquake

Scripts:
CameraControls.cs - made the reference to the camera the parent's parent instead of just parent since I added the camera shake object in between on the prefab
CameraShaker.cs
MeteorAbility.cs
EarthquakeAbility.cs


Swiftshadow:
The shake feels a bit weird, but I can't explain it better than that - probably just a me thing
Going forward, camera shaking can be abstracted to a channel
    Allow many thing to call it without needing a hard link
    For post VS, and then as a low priority
Looks good otherwise!

PR #214, opened 2021-11-07
Art Update Cloud/Wind Shader and Mushroom

Clkerr:
This pull request updates level 1 and tutorials to have cloud shadows over the level along with foliage now sways in the wind. Mushroom hazard also added for the ball to bounce off of. The mushroom has an idle and bounce animation. Bounce animation is designed for the scenario the ball hits it.


Swiftshadow:
Clouds be lookin nice!
Added a commit that will make the checkpoint text able to be clicked through

PR #216, opened 2021-11-07
Made lock go transparent instead of gray

brocEdson:
Made the lock UI icon become invisible instead of gray when not locked on to the ball


Swiftshadow:
Change made and works as per Shelby

PR #219, opened 2021-11-07
Bc2 668 hippo/ankylo animation delay

BenSchu438:
Hippos and ankylos now go through a delay first before starting, allowing for them to be synced or unsynced. Did initial value sets for the hippos based on feedback. Also did some cleanup regarding hippo prefab (removing older model parts that somehow were left over).

Affected scripts:
-HippoController
-AnkyloController


Swiftshadow:
Hippo and Anklyo delays will make the level much more interesting!

PR #220, opened 2021-11-08
Bc2 651 vs adjustments

tcaron1024:
Misc. level changes, added coroutine to make ball slow faster while on the ground, and Fred 👀 

Script Changes:
BallController.cs
StandardPlacement.cs - this is just setting the decal inactive when ability wont be able to be placed on sideways surfaces


tcaron1024:
Also allowed abilities to be placed on hazards and the large rock terrain pieces - try this out and tell me if it doesn't work/looks really jank. There will be some jank just because the surfaces aren't flat, but shouldn't be too bad for the parts that are able to be placed on

Swiftshadow:
Scene lost due to merge conflicts, backup made for redoing
Rocks and Fred moved over during review

PR #223, opened 2021-11-08
Bc2 576 Database - Level, Session, Disaster, and Machine entries

geekguy100:
##Why this is being pull requested.
With everyone continually modifying scenes, scripts, etc., that I also need to be working in to make sure data is being tracked per-scene, and with the database code being in a good state, I thought it would be best to solve any remaining merge conflicts then submit a pull request now that I have a lot of new GameObjects and script in the scenes people keep modifying. If this is unadvised, feel free to disapprove this pull request and let me know.

### The following four datum are tracked and able to be submitted to the database:
- Level
- Session
- Disaster
- Machine

### Important notes:
Level:
- A new row is added to the Level table _each time_ the player plays a level. This means that if the player plays level 0 (tutorial 1) three times, there will be a total of three entries with a "_lvl_id_" of 0 and a "_session_id_" of _XYZ_.
- As of now, no score is being tracked. Therefore, every score entry is 0.

Session:
- Thinking more about it, tracking "average disaster use time" and "average disaster hit time" makes no sense. How are we defining the "time" part of it? Is it the disaster's lifetime? In general, I think these two datum can be scrapped. Instead, we continue keeping track of use count, hit ball count, and whiffed ball count for each individual disaster. With this in mind, "_avg_disaster_use_" and "_avg_disaster_hit_time_" will both be 0 in the Session table.
- "_tutorial_view_time_" in the Session table is the time the player has spent in the tutorial levels.
- "_total_menu_time_" is the total amount of time the player is on the main menu, level select screen, or the pause screen (including instructions).

Disaster:
- A new row is added to the Disaster table _each time_ the player plays a level. For example, if I load into level 2 (first level, not including tutorials) and select Earthquake and Singularity, a new row will be added for METEOR, WIND, EARTHQUAKE, and SINGULARITY with _"lvl_id"_ = 2. If I were to replay that level and select the same abilities, completely new rows would be added the the Disaster table. I'm using an auto incrementing primary key to allow for this behavior. If this unadvised, please let me know. I thought it would be beneficial to see how players change their play styles across multiple playthroughs of the same level. If we wanted to get totals, (i.e. total num of hits for METEOR) we can write a query!

- "_lvl_id_" in the Disaster table is not a foreign key even though it should be. I'll be upgrading the current IDatabaseSubmission Queue into a PriorityQueue in order to make sure any level data is submitted before any disaster data, since disaster data is dependent on pre-existing level data.

Right now, level data is being submitted whenever the level is unloaded. This means that if a player doesn't complete a level, only a level start time will be recorded. It may be beneficial to add a "_level_play_time_" column into the Level table to see how long the player plays the level until they give up!

Right now, disaster data is being submitted when the player wins the level (gets the ball in the hole). I plan on changing this in the future, but for now this was the best approach I could take because I was having trouble with the List of active abilities being destroyed before I could submit the data when the level was unloaded.

EDIT: I forgot to add that data is _**NOT**_ being submitted automatically yet. If you want to properly submit data, make sure that A) the game is running, B) you are on the main menu (optional but recommended so you know all level data was enqueued), and C) you select the DB GameObject and left click the "Push Queue" button in the Inspector.


Swiftshadow:
Looks good, its a very good start and architecture!
Added XML comments in:
    DisasterEntry.cs
    IDatabaseSubmission.cs
    LevelEntry.cs
    MachineEntry.cs
    SessionEntry.cs

PR #224, opened 2021-11-08
BC2-576 Manager Prefab Update

geekguy100:
Added LevelManager to Managers prefab, removed lone LevelManager's from scene.


geekguy100:
Just added in camera analytics submission!

Swiftshadow:
Camera entry looks solid!

PR #225, opened 2021-11-09
Bc2 664 sticky flytrap programming

brocEdson:
Made the ball stick to the Venus Flytrap on snap so it doesn't get launched as easily, and also made the initial animation slower so it doesn't launch as high


Swiftshadow:
A bit wonky with the merge, but the scene still seems intact and all functionality there

PR #227, opened 2021-11-09
Bc2 678 implement venus fly trap idle animation

tcaron1024:
Venus fly traps have idle animation with offset. Also Fred loops. Includes scene changes but they are just setting the offsets differently for each flytrap, so not important if lost

Scripts:
VenusFlytrapController.cs


Swiftshadow:
Animation looks nice
Fred

PR #228, opened 2021-11-09
Bc2 672 spine motivator

BenSchu438:
Implemented the **_spine slide_**. Added some boxes along the spine that add force to the ball and tuned it to the designers specifications. Also found a bug on the spine collider which is why the original physics material did not work (there were 2 colliders on it, and only one had the material). Removed the non-material one.

Slightly increased size of the audio triggers as they were a bit small and sometimes wouldn't activate. 

Also fixed a bug with the venus flytrap that would cause it to call the death channel several times. 

Affected scripts:
-HippoWind - made the force an editable field. The Hippo still functions the same.
-VenusFlytrapController - Fixed bug regarding it calling death trigger several times.


Swiftshadow:
Motivator make the spine slide much more fun
    Xylophone noises actually work!

PR #229, opened 2021-11-09
Bc2 576 database logs deaths

geekguy100:
Death tracking is now in game!
Each death in the database is unique. This means that even if a player were to replay a level multiple times, each death would have its own row in the database.

A cause of death is tracked as a string in the database, but in game it is represented by an enum containing two values:
- OOB to represent the player dying from going out of bounds (i.e. lands in the water).
- HAZARD to represent a death cause by a hazard.

The last_hazard_hit column represents the unique hazard that the ball last hit. This is stored as a string and is represented by the hazard's gameobject's name (see documentation for more info).

The last_disaster_hit is the name of the last disaster that hit the ball. This is set in each ability's AbilityUsageContainer ScriptableObject (see docs below).

Let me know if you'd like more information that what has been provided.

[Click here for the documentation](https://docs.google.com/document/d/1qewrkIRpZSSy21XJa9IsSbcD6b4knQGGrHHW2I2nNq0/edit)


Swiftshadow:
Very nice documentation!
Checked the database, death entries are there!

PR #232, opened 2021-11-09
Updated Camera Controls per Shelby's and Charli's request.  

JesseCashmer:
Please be sure to reread tooltips on canUseFreeCam and min/max coordinates before reporting bugs.  CameraControls was the only adjusted script.


Swiftshadow:
Ballcam lock does not dissapear when in freecam
    I guess you are still in ballcam?
Otherwise, looks good!

PR #235, opened 2021-11-10
Bc2 163 start screen/options menu revisions

CaptainVidia:


Swiftshadow:
Use the main project to work in
    The import was clean this time, but very clearly missing pieces
Test your scenes before pull requesting
    A lot of the errors and corrections I had to make could have been very easily caught with a quick test before PR'ing
On your canvas, set it to "Scale with screen size", not "Contant Pixel Size"

PR #237, opened 2021-11-10
Bc2 693 flythrough update

tcaron1024:
Finalized flythrough of level 1. Small adjustments to level 1 (ramp not being flush to ground)

Scripts:
VenusFlytrapController.cs - setting up to play animation from flythroughcontroller
FlythroughController.cs
Checkpoint.cs - - setting up to play vfx from flythroughcontroller


Swiftshadow:
The new flythrough is very nice!

PR #238, opened 2021-11-10
Updating timer placement/display

tcaron1024:
Updated timer display/placement
Scripts:
GameplayUIBehavior.cs - changed timer to display as 00:00:00


Swiftshadow:
New timer lookin sick

PR #239, opened 2021-11-10
BC2-695 fixed bugs with wind/meteor

tcaron1024:
Made ball fire activation get handled by ball instead of meteor to fix bug with cancellation causing permanent fire. Got rid of extra collider on wind. Ball Prefab changes might conflict with stuff Ein is doing - if so, only change on Ball prefab is setting the FireEffects variable in ballcontroller to the object that is already under the Ball prefab

Scripts:
BallController.cs
MeteorAbility.cs


Swiftshadow:
Ball is now on fire for the correct time

PR #240, opened 2021-11-10
Bc2 576 database - checkpoint tracking

geekguy100:
Checkpoints are now tracked!
Each checkpoint and each ending hole have a new component: CheckpointEntry.
The CheckpointEntry is prepared for submission whenever the golf ball interacts with it, tracking the time the player took to get to that checkpoint, the time elapsed since the last checkpoint, and the last checkpoint, if any, the player was at.


geekguy100:
Hazard tracking is now in game!
Since each Hazard interacts with the ball in a unique way, there is a script on each GameObject (usually a child of the Hazard) that interacts / collides with the golf ball. This script is either `TriggerHazardEntryInteractionHandler` or `ColliderHazardEntryInteractionHandler`. Each hazard (parent GameObject) has a `HazardEntry` component which the former two scripts can call upon to increase the interaction count.

I chose this method mainly because of how the Hippo interacts with the ball - it's done through four separate triggers. Combined, these four triggers make the sucking ability work. Originally, I had a HazardEntry on each Trigger, but realized that four separate HazardEntries would be submitted. So instead, those `IHazardInteractionHandler` scripts can accept siblings, and when one sibling is interacted with, all of them shut down (i.e. prevent increasing the interaction count) for a set period of time before re-enabling.

I'll be writing more concrete documentation that covers this in more detail, but I thought it would be a good idea to summarize it here.

Swiftshadow:
Looks good
NOTE: If a computer_id exists from the old database, make sure to nuke it before submitting to the new one

PR #241, opened 2021-11-10
BC2-576 resets are tracked!

geekguy100:
New BallReset channel that is invoked when the player resets the ball. When the event is raised, the ball's position is recorded and submitted as a ResetEntry.

Since resetting the ball just calls the death event, I have a flag being set that prevents the death entry from being submitted alongside the reset.


Swiftshadow:
Its almost done! Very well done Kyle!!!!!!!

PR #243, opened 2021-11-10
BC2-696 fixed tutorial lvl 1 water collider being too high

geekguy100:
Fixes the ball colliding w/ the water's death trigger when entering the win hole


Swiftshadow:
Ya no longer die upon winning the tutorial!
Check with Charli on a more permanate solution, I know she wanted the death plane very high

PR #244, opened 2021-11-10
New ability buttons in

darlynn-best:


geekguy100:
Fixed exit button not being able to be clicked consistently. 
Updated text to reflect new controls.

Swiftshadow:
New icons are noice!
Circular cooldown also looks better than the old rectangle

PR #246, opened 2021-11-11
Bc2 679 new branch for new meteor numbers

M1sterChaos:
What I have changed for this branch is the Meteor Cooldown from 3s -> 4s, and the Meteor Max vertical Force from .2 -> .17

This was done because this makes the Meteor feel more meaningful and not “spammy”, and doesn’t feel as broken right now before rework, just for vertical slice


Swiftshadow:
Meteor balance tweaks made

PR #247, opened 2021-11-11
BC2-708 Ankylosaurus Collision update

CharlesBolt:
Updates the collision boxes of the ankylosaurs to be more friendly to the player. More work is definitely going to be needed, but this should provide a much cleaner experience overall.

Notes for the future:
Collisions for the tail need to be better, in terms of that it should be more close-fitting (and the ankylosaurs probably need more boxes to have more tight-fitting collision overall as well.) We made the collision for the tail huge because of fears that the ball would roll right past the tail, but this "solution" causes problems since it doesn't appropriately communicate with the player. The collision boxes need to follow the tail more precisely.


Swiftshadow:
NO_AUTOBUILD
Minor hitbox changes, but nothing breaks!

PR #248, opened 2021-11-13
Meteor Visual Update, decal conversion for meteor and black hole

SpencerStilwell:
Meteor got a visual update, and meteor and black hole now use decals for their projections. Meteor animation speed was also sped up by about 2X, as a request from the disaster team.


Swiftshadow:
New meteor is freaking awesome

PR #249, opened 2021-11-16
Bc2 712 reworking wind

tcaron1024:
Changed DragDirectionPlacement (used by wind, earthquake, and lightning) to show the decal on top of the ball always, then when the player clicks rotate and spawn on top of the player.  Changed wind to be scalable and use a one-time force on the ball. Fixed bugs with above changes.

Changed scripts: 
DragDirectionPlacement.cs
WindAbility.cs
WindScaling.cs
WallScaling.cs


Swiftshadow:
Styling looks good
Bug: Starting to drag over the water leads to the rotation being locked to one direction, and weird
    In general, the dragging does not work over the water, which can make mid-air adjustments difficult
Other than that bug, looks good! Get it fixed, and its on main!

PR #251, opened 2021-11-16
Bc2 579 options in pause menu

QDonnelly-254:
Options button appears in pause menu


Swiftshadow:
Will be getting merged by extension with ui-main. Review will occur then

PR #252, opened 2021-11-16
Bc2 709 update wind texts to match rework

M1sterChaos:
Updated wind numbers to deal with balancing issues, updated wind text to reflect wind rework.


Swiftshadow:
Wind numbers tweaks, yay!

PR #253, opened 2021-11-16
More Meteor Visual Overhaul

SpencerStilwell:
Sky darkens, impact decal is placed, and Fire starts when the meteor impacts. Last touches to the meteor for now, but it looks hugely pog.


Swiftshadow:
The sky darkens, the meteor falls, and the VFX explode!
How hard would it be to do variations with fewer particles in case of performance?
    Not having an issue right now, but with more VFX/particles incomming, better to be prepared

PR #255, opened 2021-11-17
Bc2 744 meteor decal arrow color changing

tcaron1024:
Meteor decal arrow and line change color based on how much force will be applied to the ball - unsure of how i like the effect (it kinda showcases how janky the backend force calculation is)

Scripts:
MeteorUtilities.cs
MeteorDecalTrajectory.cs


tcaron1024:
Hold off on this one, waiting on designer feedback

PR #257, opened 2021-11-18
Bc2 316 start screen redux

BenSchu438:
Updated TitleScreen4.0 with the new visuals and according to Shelby and Scottie. It now has a wooden sign type of aesthetic. The functionality is the same as the previous menu. Instead of a fast sliding screen, it is a static image that will fade transition between various images (currently 2 Fred and 1 David images) about every 10 seconds. 

Affected Scenes:
-TitleScreen4.0 (also moved scene to UI scene folder)

Affected Scripts:
-OptionLoadingMenu (new) - Script that allows the option menu to be opened by scenes without the GameplayUIBehavior script.
-ImageCycler (new) - Script that switches between two image objects that constantly fade in and out in a way where it looks like an image fade transition.


Swiftshadow:
Menu interactions a bit jank, but they are works in progress
Added XML comment on ImageCycler.cs

PR #258, opened 2021-11-18
BC2 131  tutorial level grayboxed

CaptainJ61:


Swiftshadow:
Had to do some level select scene changes to add the new tutorial to it
Re-added the restart button to the tutorial 2 v2 level

PR #260, opened 2021-11-18
Bc2 744 meteor decal arrow color changing

tcaron1024:
Resubmitting because Comrade Andrew wants David to see it tomorrow

Meteor decal arrow and line change color based on how much force will be applied to the ball - unsure of how i like the effect (it kinda showcases how janky the backend force calculation is)

Scripts:
MeteorUtilities.cs
MeteorDecalTrajectory.cs


Swiftshadow:
Colors are pretty
Comrade Doug is pleased
How does the color expose the jank behind the scenes? It looks pretty good to me

PR #261, opened 2021-11-18
UI main

BenSchu438:
Several UI changes have been made. 

Most work has been done to start adding in the new wooden aesthetic but there have also been various bug fixes and function changes within the scripts. Unsure exactly the amount since there have been tons of changes made in this branch.


Swiftshadow:
Cooldown circles are offset from their icon
Otherwise, looks solid!

PR #262, opened 2021-11-18
Bc2 503 level 1 jungle (revision) adjustments

IcarusPanda:
I added in the alternate paths which feature sneaky mushrooms and hippos


Swiftshadow:
The hidden hazards on the new routes are interesting
    Definitely make the player react to them instead of plan around them
    Difficult air control makes the alt routes harder, but that should be getting resolved soon

PR #263, opened 2021-11-18
Bc2 576 database submission

geekguy100:
Database data is being submitted! Hurray!

I came to the realization that pushing the data in bulk was too risky, not to mention difficult to figure out. I was trying to use C#'s WebRequest class, but I couldn't get the WWWForm data (from Unity) to send properly.

So I decided to have data submitted periodically throughout the play session **_and_** whenever a new scene is loaded.
By default, I have it set to any enqueued data is submitted every minute. If a player is playing a level, the following data is enqueued at the following times:
- Level data: Enqueued every 0.5 minutes (30 seconds).
- Hazard data: enqueued every 0.9 minutes (about 54 seconds).
- Camera data: enqueued every 0.9 minutes (same as above).
- Disaster usage data: enqueued every 0.5 minutes (30 seconds).
- Session data: push whenever a submission occurs.

Submissions to the database occur every minute.
As previously mentioned, a submission / push to the database occurs whenever a new scene is loaded (expect the fly though scenes and menus).
Upon pressing the "Exit" button, any remaining data is submitted. This should only be session data unless there is anything that is being submitted in the moment the button is pressed. The game will not exit until all data has been submitted. Granted, the player can press the "X" button (if they're not in full screen), but this is why I've prioritized periodic data submission.

I also made my timers more precise. Inspired by your GameManager's timer that ticks every 1ms, mine do the same as well (total play time, time spent in menus, tutorial times, etc.)

Finally, I have all database output being logged to a text file. It's mainly for debugging purposes to make sure stuff's getting submitted properly in a built game. These logs can be accessed in "LOCATION OF .EXE"/Disaster Golf_Data/Logs/Database/.

**EDIT**: Right now time spent in ball cam is only being tracked, and even if you're in free cam it continues to track it as time spent in ball cam. I haven't touched any of the channels, so I'm assuming something broke when the camera controls got changed again? I'll talk to you more about this in class.


Swiftshadow:
First-time bootups offline may lead to tracking issues, otherwise it works very well!
    Offline first time bootup is an edge case we can worry about later

PR #265, opened 2021-11-18
Made animation timing event-based and not hard-coded

brocEdson:
Made animation timing event-based and not hard-coded


Swiftshadow:
Sound on the flytraps no longer works
    Check with Matt what needs to be done to get that working again

PR #266, opened 2021-11-18
Bc2 594 fix camera bugs

JesseCashmer:
Main changes are in CameraController.cs.  Made newPos into a gameobject that goes around to check if it is valid to move there.  Currently doesn't do much but said bug is held out until after playtests determine which camera style to go with.

Gameplay Controls got changed to have the W/S move up and down option for the Camera Angle.  CameraController.cs got updated to reflect it.

CollisionCheck.cs is used with newPos to check if object movement is valid, currently dormant as to not cause bugs


Swiftshadow:
Changes look good!
Not doing much atm, but thats because they are dormant until post-playtest
An option to invert the W/S angling would be nice in the future

PR #267, opened 2021-11-18
Tutorial 3 Death Pits

IcarusPanda:
Changed Tutorial 2 so that it is only 3 islands, forcing the player to use disaster combinations to get across.


Swiftshadow:
Re-named and added tutorial 3 to the new level select screen
Tutorials look good!

PR #268, opened 2021-11-18
BC2-755 Direction placement based on ball position

tcaron1024:
Changed direction placement to be based on difference from ball position not click release position - changed wind scaling to work from ball position as well (cheated because the ball is always at the direct center of the screen but it works)

Changed scripts:
DragDirectionPlacement.cs
WindAbility.cs
WindScaling.cs


Swiftshadow:
Wind behaves much the same, so the under the hood change as worked!

PR #272, opened 2021-11-18
Bc2 767 venus flytrap vfx

SpencerStilwell:
Name is off because I had to rename it to BC768, it's the same Venus flytrap VFX though.


Swiftshadow:
Snap, the flytrap wind goes whoosh

PR #273, opened 2021-11-18
Bc2 759 in air placement system

geekguy100:
Refactored drag direction placement to work as expected in the air:
- The mouse plane now follows the ball around, regardless of whether or not it is in the air.
- Fixed bug in DragDirectionPlacement.cs where ball's position was not being stored properly.

NOTE: This pull request does **_not_** include the proposed "virtual joystick" that pops up when the player tries to use a disaster in the air.

Please don't delete this branch.


Swiftshadow:
Looks good!
Had to change meteor, earthquake, and singularity to not be on water, but that was very simple to do

PR #274, opened 2021-11-19
Bc2 692 start balancing geyser and lightning

M1sterChaos:
Changes cooldowns of abilities. Lightning bolt goes from 7s to 3s, wind goes from 4 to 2.6s, singularity down from 7 to 3.8s, and geyser from 5s to 4s


Swiftshadow:
I am questioning some of the numbers, but it works
    Singularity faster than geyser?
    Lightnig to 3? Its a lot faster now, and with the air placement actually being relaible, can become very OP very fast

PR #277, opened 2021-11-20
BC2-783 drag direction placement decal follows mouse before clicking …

tcaron1024:
…and wind scales

Changed DragDirectionPlacement.cs to have the decal rotate with the player's mouse before they click down. WindScaling.cs changed for the same reason - wind will scale based on mouse position before the player clicks down

Scripts:
DragDirectionPlacement.cs
WindScaling.cs


Swiftshadow:
These changes make it much easier to use the wind and lightning!
GameplayUI Updating has its cameran enabled, meaning tutorial 1 looks a bit wonky
    Scene is locked out, but its an easy fix when no longer locked out

PR #279, opened 2021-11-21
Bc2 247 geyser changes

tcaron1024:
Geyser animation now activates when ball passes over it in the air as well. Separated geyser collisions into two separate colliders - one to activate the animation and one to apply force to the ball. The force collider grows over the course of the geyser animation, so it should affect the ball roughly in time with the animation of the water coming up. Zeroed out the ball's y velocity when it gets hit by the geyser force to allow it to look more consistent - note this means that it is now extremely easy to survive forever over the water by just spamming geysers under yourself. 

Removed the wall ability from the Managers prefab and the AbilityManager in Level 1 (none of the tutorials use wall right now, so didn't need to remove from those).

Script changes:
GeyserAbility.cs
GeyserCollision.cs


Swiftshadow:
New geyser feels much better
Staying over the ocean forever is not necessarily a bad thing

PR #280, opened 2021-11-22
Bc2 788 lightning arrrow black hole visual update dropshadow tweaks

SpencerStilwell:
Lightning arrow now exists for visual clarity. The black hole has much more polish on it. The drop shadow on the ball is more visible. Slightly changed the water material. Ball X-Rays through walls.


Swiftshadow:
Lighning arrow makes that disaster much clearer
Singularity polish pass also looks very cool
Water color change is nice, but personally I prefered the first color
X-ray ball is a good placeholder for the X-ray camera, and for places we don't want X-ray camera

PR #281, opened 2021-11-22
BC2-791 Remove Free Camera

JesseCashmer:
Removed all Free Cam code in CameraControls, small change in respawn manager that became obsolete.


Swiftshadow:
ITS GONE! FREEDOM!!!!!!!!!!!!
If freecam comes back, it can be done in a better fashion
    This is purely hypothetical, there have been no talks about it returning, and likely won't be

PR #282, opened 2021-11-23
Bc2 738 proper flytrap animation syncing

brocEdson:
Made animation timing based on Animaton Events instead of being hardcoded so that it is easier to edit.


Swiftshadow:
Sound now works, and the flytrap can be mapped to an animation much more easily!

PR #283, opened 2021-11-23
Bc2 759 revamped in-air placement system

geekguy100:
## Revamped the in-air placement system (for DragDirectionPlacement disasters only, since these are the only disaster we agreed on top be placed in the air).

Instead of relying on the mouse plane GameObject, I discovered a method to simulate a Plane in 3D space - it's a struct of the type Plane. 

image

When the ball enters the air, a new method, ScreenSpaceRotate(), is called instead of RotateDecalToMouse(), which is now only called _before_ the ball leaves the air. Once the ball leaves the air, even after it lands, the new method is used to get the direction to rotate the decal.

I'm raycasting onto this plane from the screen center and from the mouse position. If both rays hit the plane, the direction between the two is calculated. This is the direction that the decal will rotate in.
If the rays do not hit the plane (i.e. the camera is parallel to the ball, and the rays cast into the horizon), I calculate the forward direction by calling Vector3.ProjectOnPlane(mouseRay.direction, Vector3.up). This returns the forward direction the decal should be rotated in. 
*This method works well for a situation like this, but not well when the camera is not parallel with the ball. Just an FYI as to why I'm not using it for everything!

Finally I set the decal's rotation by calling Quaternion.LookRotation(forward, Vector3.up).

## Why not just use the old mouse plane?
It's nice to  not have to rely on a GameObject that we have to resize, scale, place into the game world, etc.
Honestly though, after several iterations of getting the decal to rotate without raycasts, this was the method I came up with; however, the behaviour is quite similar to the old mouse plane's. The main improvement is the addition of the "if camera is parallel to the ball, use Vector3.ProjectOnPlane()", However, this implementation is something we may want to test the performance of and consider using down the line. I think it's best to keep it for now.


Swiftshadow:
Placement behaves the same, which was the goal
When rotating with middle mouse button, it does spaz
    Likely a result of how the two points are drawn between when the cursor is in the middle of the screen, and the screen is moving

PR #284, opened 2021-11-23
Updated wind numbers

M1sterChaos:
Increases min force from 500 to 600 and reduces max force from 4250 to 1800 to better match the decals as play testers got confused


M1sterChaos:
Revoking the changes

PR #287, opened 2021-11-23
Bc2 781 win screen UI update

BenSchu438:
Ignore the name of this branch as Quinn did that work in UI main and I took over this branch for my UI backend rework. Made the Gameplay UI Behavior script a bit easier to read and work in as I uncoupled it from a few things, primarily the golf hole manager script. 

Affected Scripts:
-GameplayUIBehavior - Lots of changes here, mainly adding in a new method that acts like a constructor and edits the script to make said constructor work
-LoadingBehavior - Reworked to take in various variables and then pass those into GameplayUIBehavior once the UI scene loads, allowing the constructor to work
-GameManager - Added minor part to pause menu to make sure it doesn't restart the timer when unpausing while in a menu
-GolfHoleManager - removed scene determining code, as that is now contained within GameplayUIBehavior/loading behavior. This script might now not serve much of a function anymore.
-Ball Controller - Added reference to detect when the ball first gets hit by an actual ability so the timer only starts once the ball is hit
-AbilityChoiceController - Commented out its refernce to game manager to start timer once the select menu closes.

Affected scenes:
-GameplayUIUpdating - Wired with the new references as well as added the new par time stuff
-Tutorial 1 - Updated its loading behavior to properly load in the new UI
-Tutorial 2 - Updated its loading behavior to properly load in the new UI. Now uses GameplayUIUpdating instead of timer test!
-Tutorial 3 - Updated its loading behavior to properly load in the new UI.
-Level 1 - Updated its loading behavior to properly load in its new UI.


Swiftshadow:
Pulled into ui-main
Fixed bug where pausing and unpausing during the win screen caused the next level to start frozen
    Fixed by moving the Pause and Unpause functions in the GameManager out of there, and unpausing when the next level is hit, and setting timescale to 1

PR #289, opened 2021-11-23
Art main

EinEstes:
Xray Camera. Procedural mossy Rock materials, more animations for ankylo, fred, and archeoptryx.


Swiftshadow:
X-ray shader lookin pretty cool!
    Allowing certain objects to be x-ray-able and others not is a good choice
    Also meshes well with the orange see-through-walls material/shader
Earthquake did bork, TJ did fix

PR #291, opened 2021-11-24
Art main

EinEstes:
Ribs and rock not see through. Mushroom stem not defunct magenta


Swiftshadow:
Nice quick fixes before R*

PR #295, opened 2021-11-30
BC2-833 turning off hole collider when player enters to prevent enter…

tcaron1024:
…ing multiple times

Added one line to GolfHole.cs - turning off collider


Swiftshadow:
Simple fix, it works!

PR #296, opened 2021-11-30
BC2-835 fixing meteor bug and cursor bug hopefully

tcaron1024:
Fixed meteor affecting ball after its explosion ends, and hopefully fixed cursor staying as the last disaster used when going to the menu after winning sometimes. 

Scripts:
MeteorAbility.cs - new function to turn off collider when explosion ends (called by animation event)
CursorManager.cs - added line to OnDestroy that changes the cursor back to the default when this manager is destroyed


Swiftshadow:
Looks good, none of the bugs mentioned were encountered

PR #297, opened 2021-11-30
Art main Tutorials Level Comp

awielgo:
Asset placed in all tutorials, pulled main into art main


Swiftshadow:
New asset placement looks solid!
    The trees on tutorial 3 near the hole are a nice touch

PR #302, opened 2021-12-02
Hazards Changes in Level 1

RobertLee-bradley:
Added Pterodactyls on Left, removed pterodactyls from middle and right path, added mushrooms on middle path, added anklyosaurs on right path.


Swiftshadow (approved):
New placement looks solid!
Keep in mind the anklyos are being changed to proximity-based activation

PR #305, opened 2021-12-03
Many Things

Mjsciarabba:
-Bone slide notes
-Meteor, Lightning SFX updates
-Small flythrough change


Swiftshadow (approved):
Meteor dropping SFX was startling the first time
    Its great, but I was not expecting it
Thunderclap when using lightning is awesome
Merge conflicts caused the spine slide notes to require re-adding
    This has been done
    Going forward, make sure to update from main and check the lockout sheet before doing any scenework

PR #307, opened 2021-12-05
Bc2 839 fixing up tutorial boxes

QDonnelly-254:
Added new tutorial boxes for level 1 as well as fixing up the disaster select screen


Swiftshadow:
Merged into ui-main
New disaster selection looks solid!
Pause menu turned on by default - resolved
Timer appeares on all scenes, does not count unless that scene has timer enabled

PR #309, opened 2021-12-07
Bc2 723 ankylosuar test level

RyanSpringer1500:
Submitting the proximity Ankylo perfab, along with its trigger prefab.


tcaron1024 (approved):
Seems to work, but the ankylos in the current level 1 aren't prefabs for some reason so those ankylos will need to be replaced
Not sure the purpose of the ankylo trigger prefab, seems to work fine without it

PR #310, opened 2021-12-07
Bc2 843 tutorial one text edits

QDonnelly-254:
not 100% bug free-some tutorials skip for some reason, but the feature is there. will bug fix in the coming days


Swiftshadow:
Merged into ui-main
Aside from the bugs mentioned, looks solid!

PR #311, opened 2021-12-07
Bc2 804 implement archeopetryx

brocEdson:
Implemented flocks of archaeopetryxes (archaeopetri?) flying through the sky


Swiftshadow (approved):
Archie's look good!
Had to re-add the spawner to the scene, but that was a painless process.
Next steps are getting the grow-in working, and making sure they spawn off-camera

PR #312, opened 2021-12-07
UI main

QDonnelly-254:
Tons of UI overhaul stuff! Will bugfix as needed in the coming days


Swiftshadow (approved):
Said bugs from above merge into ui-main can be resolved during the bugfix time
Tutorial 1 popups cycle when the pause menu is open
Love the new disaster select and hotbar!
    Removing the videos in selection will reduce how much the player knows about each disaster before picking

PR #315, opened 2021-12-08
Bc2 846 option menu redux

BenSchu438:
Various changes to final UI tweaks, namely options menu changes. Note that the option menu is still having issues with the audio changes staying when re-opening the options menu, but that bug will be addressed during the bug fix period. Also added the new game play videos for help as well as adding to the ability data objects.

Main Scene Changes
-OptionPauseMenu - New assets added, most functionality added. More bug fixing needed. 
-GameplayUIUpdating - Added video frame, adjusted scaling of a few objects. 
-InstructionsMenu - Fixed a minor issue with the sizing of the left buttons.
-LevelSelect - Removed Sound Manager [Part of option menu bug]
-TitleScreen4.0 - Removed sound manager [Part of option menu bug] 

Script changes
-SoundManager - Added more functionality for muting audio [still need bug fixing]. Added 'Revert Changes' function, though has issues with mute audio.


Swiftshadow (approved):
Options menu now fits the rest of the theme!
Thank you for noting the bugs: Fixing them is what the bugfix time is for
    The revert to default button did not seem to work (already listed for sound, not sure if the rest is known)
SoundManager will need to be brought up to standard, but atm the new visualts take priority

PR #316, opened 2021-12-08
Bc2 831 new flythrough

tcaron1024:
New flythrough level, renamed to be the same as the old one so it should still connect up to the UI. Minor code changes to FlythroughController.cs to allow for more sounds to play.


Swiftshadow (approved):
Flythrough looks good!
Hopefully the build settings tweak will let the game get built automatically as well
Archie animation controller contents got nuked, re-added that in

PR #319, opened 2021-12-09
Bc2 851 fixing flythrough

tcaron1024:
Turned off disaster video audio


Swiftshadow (approved):
Tutorial videos no longer have sound - they play correctly!

PR #321, opened 2021-12-09
UI main

darlynn-best:
UI Updates


Swiftshadow (approved):
Love the new disaster select!
Lots of bugfixes too, thanks for finding them!

PR #322, opened 2021-12-09
BC2-855: Short and dumb thing here, literally just adjusting a text box

QDonnelly-254:
adjusting tutorial 1 "select 1 key" text box to (hopefully) make it so the text appears fully


tcaron1024 (approved):
Good change

PR #328, opened 2022-01-20
Fixed text bleedover on disaster select

QDonnelly-254:
Text on disaster select used to be font size 16, is now size 14 so the text doesn't bleed over onto the wood paneling.


Swiftshadow (approved):
Merged into ui-main-2022
Minor change, no code changes
Looks good!

PR #329, opened 2022-01-20
Fixed Terrain Pieces

BenSchu438:
Fixed the spawner tool's ability to spawn terrain (formerly islands) and utility prefabs. Still working on fixing the hazard spawning and decorative spawning, if we decide we still want this tool to spawn decorative items, as I know polybrush is mostly used for decorating levels. 

Only edited the LevelPieceSpawner.cs script. As mentioned, I'm working to fix the hazards and decorating spawning, but I wanted to quickly get the terrain stuff fixed first so the level team can start grayboxing new levels ASAP. Feel free to leave this branch open for me to finish fixing it, otherwise I can create a new branch to finish fixing the tool.

To use properly in the new level loading system, its important to make sure that the scene in the hierarchy is set as the active scene, as the spawner tool will spawn the pieces in the currently active scene. Will relay this fact to level designers as well.


Swiftshadow (approved):
Modification works!
New bug noticed: If a piece is spawned via the tool, no other changes are made, and the scene gets unloaded (by opening another scene), the added object is not saved
    Spawning a piece with the level tool does not appear to alert Unity that something changed in the scene, so it does not allow a save

PR #330, opened 2022-01-21
BC2-869 removed unused, old database tracking components from the hole

geekguy100:
Refactored GolfHole.cs to no longer require old CheckpointEntry script and to ignore all database interaction whatsoever.
NOTE: If you see "GameObject has an unreferenced script" or "Missing MonoBehaviour" please note that these issues have been solved on my Database branch and will be merged into main in due time!


Swiftshadow (approved):
Bug did not appear during testing

PR #331, opened 2022-01-22
Bc2 856 database 3.0

geekguy100:
Data is being tracked properly in game and is only being submitted when playing on a build of the game.
Data is submitted when exiting a level. More specifically, it is submitted when restarting a level, returning to the main menu from a level, exiting the game from a level, or progressing to the next level.

All old scripts have been deleted. All new scripts reside in Scripts/Analytics/DB 3.0 folder.

Note: When the data is being submitted (build-only) a loading screen is currently not present, so the menu remains open. This is what I plan on working on next. We could use a general loading screen and nice scene transitions anyway, instead of a sudden jump to the next scene.

Please let me know if you'd like updated documentation. Adding components for statistics tracking has been streamlined, and all current prefabs have been updated with said components.


Swiftshadow (approved):
It finally works
And does not crash (famous last words)
Nice setup with the submission
Added some XML comments and tooltips in:
    DBEnqueuer.cs
    ResetEntryManager.cs

PR #332, opened 2022-01-22
Bc2 279 optimization work

tcaron1024:
Mostly optimization stuff - went through the profiler and found some things that stood out and cleaned them up a bit. Changed physics layer interactions so that OnlyInteractWithBall layer only interacts with ball to reduce unnecessary collisions, changed physics layer of venus flytrap for same reason.

Scripts:
MaskScript.cs - this script was generating a lot of garbage per the profiler, reduced that a bit, functionality is the same
AbilityManager.cs - fixed a bug where using the number keys to switch between abilities caused the switch to happen twice, added the DragDirectionPlacement and StandardPlacement components to the AbilityManager permanently so that they don't need to be destroyed/added every switch
IAbilityPlacement.cs - changes to get placement scripts to work with new ability switching
DragDirectionPlacement.cs - changes to work with new ability switching
StandardPlacement.cs - updated in line with IAbilityPlacement changes
VectorPlacement.cs - same as StandardPlacement


Swiftshadow (changes requested):
Base changes look good
Cannot use wind, lightning, or earthquake in tutorials
    They can be selected, but not actually used
Missing XML comment in MaskScript.cs, but the rest of the file does not follow convention either
    Passing to keep the file itself consistent internally

Swiftshadow (changes requested):
If the ball is moving fast enough, wind and lightning place in the ball's direction of motion, not the direction you clicked
    See https://drive.google.com/file/d/1R_s_QRBawktPP0muMqn3OPQbL6RygFch/view?usp=sharing
Occasinoally (unable to relaibly recreate), on T3 Geyser becomes unplaceable after using Lightning
    Best guess is due to the MousePlane not being disabled correctly, but unable to reporduce to check
The unable to use Lightning bug does appear to be gone though!

tcaron1024:
Merging in changes from Kyle's  placement branch fixed bugs with wind/lightning getting placed in the wrong direction. Moved mouseplane deactivation slightly to fix geyser placement problem. Fixed a new bug with the DragDirectionPlacement that caused problems spawning while the mouse was in the air.

Script changes (since last review):
DragDirectionPlacement.cs (bug fix inside CheckAndRotateDecal())
FlythroughGeometry.cs (just added a null check to get rid of an error when stopping the game during the flythrough)
IAbilityPlacement.cs (moved mousePlane deactivation below StopAllCoroutines to make sure it wasnt getting reactivated)

Swiftshadow (approved):
No new bugs encountered!
Added XML comments on:
    AbilityManager.cs line 96
    AbilityManager.cs line 101

PR #333, opened 2022-01-23
Bc2 868 wind ability placement malfunctions

geekguy100:
No longer using screen space rotation option due to inconsistencies.
Now solely using the MousePlane to catch raycasts to determine the direction to rotate the decal in.
If no suitable terrain or the MousePlane caught the raycast from the main camera to the mouse cursor, we get the forward direction from the the mouse ray's direction and call ProjectOnPlane. See image below.
image


Swiftshadow (approved):
The wind flipping weirdly when used on the spine slide is gone!
Encountered no other inconsistencies with win's usage (or lightnings)

PR #334, opened 2022-01-23
Bc2 870 level creation tool

tcaron1024:
New level creator tool to make it easy for designers to create all the necessary scenes for a new level. Creates Managers, Loader, Hazards, Foliage, and Geometry scenes for the given level name, creates managers/camera in Managers scene and Loader Objects in loader scene, sets scenes to load in the loader object, and adds the new scenes to the build settings. Looked into using scene templates, but decided to stick with the scriptable object for now so that designers can have the tool ready for Tuesday.

Scripts:
LevelCreator.cs
LevelCreatorScriptableObject.cs
LevelLoader.cs (added function to allow for setting the scenes to load through code)


Swiftshadow (approved):
This should make the Level team's job much easier!

PR #335, opened 2022-01-25
Art main

EinEstes:
Changes to procedural mossy rocks, dodo bird, and changes to xray lens with camera.


Swiftshadow:
Delayed due to the models not matching the texture grid

Swiftshadow (approved):
UV's be fixed
Nothing else appears to be broken!

PR #336, opened 2022-01-25
BC2-872 meteor lighting change fixed and lightning lighting change added

tcaron1024:
Fixes meteor sky darkening effect and adds it to the lightning. I created a Light Manager object and script to handle changing the lighting so that it is easier to do in the future. Added the LightManager to the Managers prefab.

Scripts:
LightManager.cs - new
LightningBoltAbility.cs - Added reference to channel that triggers the lighting change and variables for lighting change
MeteorAbility.cs - Changed code up a bit to work with LightManager (similar to lightning bolt changes)
SpawnObjectsOnMeteor.cs - Got rid of all code meant to change lighting in this script (this one is one of spencer's so comments are a bit lacking)


Swiftshadow (approved):
Lighting changes be back!
Centralizing them all in one spot is a good idea

PR #337, opened 2022-01-26
Bc2 856 database 3.0 - session ID bug fix

geekguy100:
The Session ID was being re-calculated whenever a new scene was loaded. It got messed up when my computer blue screened, but has now been fixed.
Also removed checkpoint interactable from final hole prefab - we can determine it from the level complete time already.


Swiftshadow (approved):
Fixes a session_id generation bug, so very useful!

PR #339, opened 2022-01-26
Bc2 856 database 3.0 - updated level ids with level names

geekguy100:
### Now using the name's of the level instead of the level ID.
The names get recorded in the following fashion:
`(L | T)X`,
where the first character is either **L for a level** or **T for a tutorial**, and X is the number of the level (level 1, level 2, etc.)

### UPDATE:
Updated ints to strings in the rest of the manager classes - they were still using ints!
tagged the Sneaky Hippo as a Hazard and changed its name as such for easier database tracking.


Swiftshadow (approved):
Level names are much easier to understand now!

PR #340, opened 2022-01-26
Bc2 863 fix spawner tool

BenSchu438:
Various changes to the Spawner tool to properly work after the recent file changes. Below is a complete list of changes. 

Affected Scripts:
-LevelSpawnerTool.cs / ObjectSpawner.cs (renamed)

Changes:
-Renamed tool to Object Spawner tool. Did this as its catchier and to avoid confusion with the new Level Creator tool which creates the scenes
-Fixed bug where the scene wouldn't be marked as 'Dirty' after spawning an object
-Renamed 'Functional' to 'Terrain', as its mainly used for spawning the islands
-Moved Hazards to its own category within the tool, with its own subcategories Hippo, Ankylo, Ptero, Archie, and Uncategorized
-Tool now automatically grabs prefabs with the 'LevelPrefab' keyword from within the Hazard folder in resources. Any objects with that keyword and another keyword, such as Hippo, is auto categorized under Hippo. Any objects that don't have a category keyword will be placed under the Uncategorized menu. 
-Added the LevelPrefab keyword to all existing level-ready hazards. This keyword should only be put on hazard objects that have its model, animations, effects, and scripts, as it means that prefab is ready to be in a level. This is used to distinguish it from the models and animations within the tool. 

Looking into better optimizing the code and expanding hazard's functionality to the terrain, but that can take some time, so releasing this so the level team can use the hazards.


Swiftshadow (approved):
Looks good!
Would it be possible to make the prefab keyword read in via file?
    In case it changes, or new keywords are needed, a script change would not be necessary
    Not at all mandatory, and a very low priority if you do want to

PR #342, opened 2022-01-27
Bc2 857 vertical level geometry

RobertLee-bradley:
Has grayboxes for 2 new levels, mushroom level and curvy bridge.


Swiftshadow (approved):
Solid base for the new levels!
    Big mushroom go boing
Issues fixed:
    DeathPlane not actually killing the player
    DeathPlane not allowing disaster placement
    Timers on 2 new levels starting immediatly, not on first disaster hit

PR #343, opened 2022-01-27
Bc2 875 level jungle hippo geometry

NatSmarant:
A graybox of a level where you must escape the sucking of a hippos mouth and make it to the hole.


Swiftshadow (approved):
BEEG hippo
Had to move the ball starting location to avoid the timer starting early

PR #344, opened 2022-01-28
Bc2 882 maze greybox

tcaron1024:
Maze level greyboxed. Added teleport hole prefabs to make maze functional - these are in the UtilityObjects folder 

Scripts:
TeleportHole.cs - new
PterodactlySpawner.cs - fixed bug causing pterodactyls not to spawn when ball entered area


Swiftshadow (approved):
Interesting level
Not sure how well it will work, but its definitely a good thing to try

PR #345, opened 2022-01-29

M1sterChaos:
Finished preliminary mushroom level
Made script in mushroom level to "move ball"
May need more later to make the level feel better but my tests worked

Please don't delete branch after pull request, I still need to finish the other 2 levels


Swiftshadow (approved):
Only the mushroom level set up to the level select
Even more interesting than TJ's level
Its ready for playtesting though!

PR #347, opened 2022-01-30

M1sterChaos:
Added windmill level
Added script to control speed of windmill
new folder that holds the script that controls the windmill

Please do not delete this branch after merge, I have not finished the third level but I want to brainstorm it more


Swiftshadow (approved):
Please put hazards in the Hazards scene
    If using the level tool, right-click on the hazards scene and do "Set as Active"
        The tool places in the active scene, so this will place hazards in the hazards scene
Used this branch/PR for some other housekeeping
    Level names no longer include spaces (except for the suffixes
    Removed some leftover text in the Level Select scene

PR #348, opened 2022-01-30
BC2-887 basic version of water hazard - ball dies after being over water too long

tcaron1024:
Ball now dies after staying over water for longer than 12 seconds. Once the model/animation is done for the water hippo, the balldeathchannel channel trigger can be changed to trigger a channel that starts the animation/whatever we end up with for it. Might also be good to have some UI indication that the ball needs to get back over land, but that can be added later. Not sure if you want to wait on that stuff to merge or not.

Scripts:
BallController.cs - adds coroutine to raycast and check if the ball is over water and a coroutine to countdown the time the ball is over water and kill the ball


Swiftshadow:
Delayed until some visual indicator is put in

Swiftshadow:
Maps a very rough shark jump anim to the over-water death

Scripts:
SharkController.cs (new script)
TransformChannelSO.cs (new script)
BallController.cs (changes to how/what the over-water does)

Swiftshadow (approved):
Scorpion looks good
Functionality works well
Above from TJ

PR #349, opened 2022-02-01
Bc2 873 loading screen and database submission fixes

geekguy100:
- Fixed empty submissions for computer ID.
- Loading screen that pops up when the data is being submitted.
    - Input is also disabled just in case when the data is being submitted.
- Refactored the GameManager to be a more sophisticated Singleton - The EnhancedSingleton class has more quality of life features, checking, etc. 
     - If a GameObject doensn't already exist in the scene of that type, the EnhancedSingleton can make a new one and make sure it is DontDestoyOnLoad()! (Although this is problematic for types like the GameManager where we need to manually drop in channels. Addressables could probably help with this).
- Fixed other bugs that popped up along the way.


Swiftshadow (approved):
Does seem to work!
Added in some XML comments and tooltips in areas missing them
Loading screen needs a UI pass, but that can come later

PR #350, opened 2022-02-01
Bc2 888 question mark rocks

RobertLee-bradley:
Added Actual Water to Mushroom and Curvy Bridge Level

Added a Big ol Rock to the Curvy Bridge Level

Adjusted the hitbox and strength of the Big Mushroom on Big Mushroom Level


Swiftshadow (approved):
Rock
Very large rock
Big mushroom bounce is a lot more noticable
Also you don't clip into it anymore!

PR #351, opened 2022-02-01
Bc2 877 level jungle jump geometry

NatSmarant:
Jungle Jump level gray boxed with working hazards and level selection


Swiftshadow (approved):
I assume there will be more than just the ptero's
    It seemed very empty, even accounting for no foliage placement
Ptero's sometimes to hit the walls and delete
Level added to the select screen

PR #352, opened 2022-02-01
Bc2 900 function changes

RyanSpringer1500:
The implementation of the immediate functions for the Hazards. Double check for coding errors just in case.


Swiftshadow (changes requested):
All except Flytrap look good
Flytrap does not work
    L1 Flythrough uses it, and the flytraps did not animate
    Triggering them similar to the hippo is not a bad idea
        Make sure to trigger the leftAnim and rightAnim objects - those are the ones with the animations

tcaron1024:
Not gonna have time for the changes needed here, closing

PR #354, opened 2022-02-01
Updated Tool with rocks

BenSchu438:
Updated the tool to properly find all the rocks available to spawn for designers. New 'SmallRock' category that finds them all based on the same rocks with LevelPrefab and Rock in its name within the environment folder instead of levelpieces folder.

Added ability to overwrite the keyword used to find objects by updating a text file that is now within the Editor folder. Editing that text file with the new keyword will take effect when relaunching the tool. If it cannot find that file on boot, it will then use the default listed in the script. 

Renamed all the rock assets to have LevelPrefab keyword within them. Added colliders that were missing from 2 rocks in the rock folders. Removed LevelPrefab keyword from quicksand as that is no longer being used. 

Affected Scripts:
-ObjectSpawner.cs
-ObjectSpawnerKeyword.txt (new - in Editor folder with script)


Swiftshadow (approved):
Looks good!
Config text file is a nice QOL addition

PR #355, opened 2022-02-02
Bc2 890 maze level iteration

tcaron1024:
Changes to the maze level:
Simplified the route (no longer possible to infinitely loop through islands)
Added signposts to each teleport hole to give the player more idea of the choice they are making
Increased succ force on teleport holes to make it less frustrating to get in them
Added rocks between islands to make it harder/less obvious that you can just skip it all with geyser/lightning

New prefab - Blank_Sign_With_Textbox (in Resources/UtilityObjects)


Swiftshadow (changes requested):
Love the addition of signs!
    The names are pretty fun
Removing loops is a great thing
Desync the hippos. Right now they are all robotically in-sync
   The hippos actually doing the motion/with the script are child objects to the prefab the tool spawns

Swiftshadow (approved):
Hippos no longer robots

PR #356, opened 2022-02-03
Bc2 934 jungle ramp level

NatSmarant:


Swiftshadow:
Delayed due to confusion on which level to merge add to the select screen. Will revisit once confusion is cleared up.

Swiftshadow (approved):
Solid level
Added to the select screen
I hope there is a lot of foliage, otherwise this will be pretty blank

PR #357, opened 2022-02-03
Bc2 950 testing camera rotation change

tcaron1024:
Changes to camera rotation - gets rid of cursor lock. Also makes the mask script get its capsule collider on start instead of being public (it got unset on main and was causing errors).

Scripts:
MaskScript.cs
CameraControls.cs


Swiftshadow (approved):
You have pleased the UI/UX God Shelby

PR #358, opened 2022-02-03

M1sterChaos:
Finished preliminary vertical level. Can make longer easily if needed.

Branch can now be safely deleted if you want to


Swiftshadow (approved):
This level is very weird, but it does function
Maybe this can become a volcano level?
    Scorpion is currently flinging the ball upwards

PR #359, opened 2022-02-03
Bc2 933 dodo implementation (non glitched)

brocEdson:


Swiftshadow (approved):
Dodo birb!
Will be placed by Ashley, not level designers
    Location is Prefabs/Resources/FinishedAssets/Environment/E Universal/EU Creatures
May need a way to delay their start in the future, can be added if requested

PR #360, opened 2022-02-03
BC2-952 rocks now prevent ball suck on large hippo lvl

geekguy100:
Added triggers behind rocks that disable the wind colliders, preventing the ball from being sucked back into the hippo's mouth. When the ball exits the trigger, the colliders are re-enabled.


tcaron1024 (approved):
Hippo suck gets stopped
Might need some visual indicator for that at some point
Added tooltips for variables

PR #361, opened 2022-02-04
Pausing disables placement and wind/hippo fixes

tcaron1024:
Pausing the game now stops disaster placement and unpausing resumes it. Also noticed that the hippo variant prefabs had some unset variables and missing scripts that were preventing them from working properly/saving (missing scripts were on the hippo wind objects so I don't think they should have been there at all) so I set those variables and deleted the missing scripts. Fixed a bug with the wind causing it to activate twice sometimes.

Edit: ability video pausing was added to branch bc2-871 to prevent merge conflicts

Scripts:
AbilityManager.cs - main pausing/unpausing disaster placement changes
DragDirectionPlacement.cs - small changes to accommodate pausing placement
IAbilityPlacement.cs - added null check
WindAbility.cs - fixed bug causing it to activate twice sometimes


Swiftshadow (approved):
Looks good!
Can the disaster disabling be added to the win screen as well?

PR #362, opened 2022-02-05
Vfx main

tcaron1024:
Big change is sandstorm now has its vfx done and implemented. It is not added to the abilitymanager because I didn't want to mess with disaster select at all, but it should be good to go when dragged in.


Swiftshadow (approved):
Sandstorm VFX make it a bit hard to see the ball
    Arrow also blends in to the dust
Disaster definitely needs some balance, but that can come now

PR #363, opened 2022-02-08
BC2-1083 Possible fix to meteor/singularity placement bug

tcaron1024:
Should fix the meteor/singularity sometimes becoming unplaceable. 

Scripts:
StandardPlacement.cs


Swiftshadow (approved):
The bug is difficult to re-create normally
    This does not break anything, so assuming it also fixes the bug at hand

PR #364, opened 2022-02-08
Vfx main

SpencerStilwell:
Fixed visual bugs on meteor and lightning. Also added new visual effects on lightning and updated meteor fires.


Swiftshadow (approved):
The lighning upgrade is pretty cool
    Its not as over-the-top as meteor or earthquake
        Not a bad thing
Did not notice any changes on the meteor
    That must mean whatever bug there was is now fixed

PR #365, opened 2022-02-08
Bc2 876 connors levels

IcarusPanda:
Transition Level


Swiftshadow (approved):
The ball is white. It will stay that way.
What the flying Fred
Love the idea, the volcano climbing section is really cool
    May be a bit long on the ladder part
This level will be interesting to screenshot
    And flythrough
Please split the models and other non-scene things to their correct folders
    If you need help with this, ask myself or Ein
Used this PR to swap the sprite-based level headers to be text-based

PR #367, opened 2022-02-08
Bc2 1081 disaster balance pass

tcaron1024:
Balance adjustments to earthquake, singularity, sandstorm. Added an outer pull to the singularity, increased earthquake forces, and sandstorm cooldown increase. Also made it so that drag direction decals (lightning, wind, earthquake) rotate better on ramps - it is still a bit wonky, but people have agreed that it is better than it currently is on ramps (the decals don't rotate with the verticality of the ramp on main)

Scripts:
DragDirectionPlacement.cs
SingularityOuterPull.cs


Swiftshadow (approved):
NIce changes overall!
Was able to get the earthquake to not fire on a ramp by ramming into it really fast (lightning-wind buildup, earthquake already active)
    Only happened once, having difficulty recreating
Fixed a typo I made in a level merge loading the wrong scene

PR #368, opened 2022-02-08
Implemented the lizard

brocEdson:
Lizard only runs away if the ball is at a certain speed, to prevent the lizard being scared of a stationary ball
Lizard works only on flat surfaces, getting it working on any incline would be a pain
There's some weirdness in how the lizard turns away from the ball when running, and I don't know what that's about. If you know what's happening, feel free to fix it but it works well enough


Swiftshadow (approved):
The lizard can bounce off mushrooms
    This is incredibly funny
Lizard occasionally got stuck in one spot
    Unable to recreate
    My best guess is it generates the same target, and so moves nowhere

PR #369, opened 2022-02-10
Bc2 875 level jungle hippo geometry

NatSmarant:
Small update to the level and the blocks stopping hippos suck is in.


Swiftshadow:
Encountered a bug with the wind cursor, investigating

tcaron1024 (approved):
Cursor bug is gone!

PR #373, opened 2022-02-10
Boost pad, stormcloud, and parallax map decals for level artists.

SpencerStilwell:
Pog? I'm gonna teach the artists how to use the parallax decals.


Swiftshadow (approved):
Made the thunderstorm VFX cloud function as the hazard
    Bolt spawning can be put in later
Boost pad VFX are pretty cool

PR #374, opened 2022-02-10
Bc2 1084 camera rotation tool

geekguy100:
Created a tool that, when the ball collides with a special GameObject, the camera will pan towards a given look point, wait, then pan back before returning control to the player.
Custom editor for ease of use - accessible via Tools/Create Camera Rotation Trigger.

Made at the request of @NatSmarant for hill level. Additionally may be used elsewhere.
image

Note: This isn't the full sequence - Gyazo cut me off cause I have the free version 🥴
1eae80846fc5a0629d015d0063a97a4a


Swiftshadow (approved):
Make sure the rest of the designers know about this tool!
I like how you did the lookpoint creation

PR #375, opened 2022-02-10
Bc2 1094 falling rocks

brocEdson:


Swiftshadow (changes requested):
The falling rock does not need the same level of sophistication as the ptero
    Ptero is a good reference, but is more complex than this needs
        Rotation or varying heights are unecessary
Falling rocks should fall everywhere in the area, not just around the player
    Starts falling when the player is in the area still

brocEdson:
Keeping some rotation code just because it'll make this look nice when we get art assets for it, otherwise made all changes requested.

Swiftshadow (approved):
Good call on the rotation
I'm pretty sure the final models (for the falling rocks) are in, those can be applied in another branch/task

PR #376, opened 2022-02-10
BC2-1096 Updated Earthquake Behavior

geekguy100:
The earthquake now pops the ball up every useRate* seconds, so long the ball is inside of the Earthquake's hit trigger. 
*Assignable via the Inspector

I think this change makes the earthquake more consistent and makes it work more as intended; it behaves more like an earthquake does. Before, when used up hills, chances are that the earthquake would rapidly pop the ball up since the ball was constantly entering and exiting the hit trigger. Now it acts on a set pulse, and will pop the ball up every useRate seconds so long the ball is inside the hit trigger.

@tcaron1024 looking for feedback. Much appreciated 😄


tcaron1024 (approved):
Earthquake functions a lot more consistently now
useRate value of 1 looked good to me but you should probably check with designers on it

PR #377, opened 2022-02-11
Bc2 836 level name button

BenSchu438:
A minor rework to the internals of the level select screen. Now uses scriptable objects that store Level Data. They contain the level name (also the name of the SO itself), the name of the loader scene associated to that level, and a thumbnail image for that level. 

Now, within the Level Select canvas object in the screen, it holds a list of LevelData objects. When loaded, the script will automatically generate and prepare a level select panel for each LevelData object. It also properly scrolls through level names at the top, and the top title is also a button that can load the level. Also fixed the button spam bug (or at least I think so, ppl with faster clicking speed can test). 

This means that to add a new level to the menu, a new level data object needs to be made, set up with an image and scene (deletes itself otherwise), and then added to the LevelData array within the Level Select script. The script will do the rest and we only need to update the scriptable object/list order to change anything we want. 

Script Changes:
-LevelData.cs (new) - ScriptableObject for level data that current holds the scene name and level image, and able to return its own name which is the level name (can be changed if desired, but I thought this would avoid any confusion on how to rename the level later)
-LevelSelect.cs - Quite a drastic rework. Generates the level panels at start function and call the current page to load its level. Updates background and header text.
-LevelPanel.cs (new) - Script attaches to the level panel itself. Takes in a LevelData object and loads its thumbnail and the name of the scene it needs to load when activated.


Swiftshadow (approved):
Adding new levels to the select will be a lot easier now
The BGM destruction can be optimized slightly:
    Cache the GameObject.Find(), and test if the cache'd var is null

PR #379, opened 2022-02-12
BC2-1241 dragonfly schmoovin

tcaron1024:
Dragonfly moves in random directions. Its prefab has moveDelay, moveTime, rotateTime, and min/maxMoveAmount variables to control exactly how it moves. It also has a moveCollisionLayers LayerMask variable to set which layers the dragonfly should look for to avoid it phasing through objects - I thought lerping the movement would be easier than giving it a rigidbody and dealing with physics, but that can be changed if desired.

Script:
DragonflyController.cs


Swiftshadow:
Delayed until Ashley gives her opinoin

tcaron1024:
Ashley approved dragonflies

Added a bunch of code to DragonflyController.cs to add more control over how much they normally rotate, the max distance they will travel from their start point before turning around, and how they attempt to get out of tight spaces

Swiftshadow (approved):
Ashely-approved
Dragonfly looks a lot less robotic
    Animations will help a lot there as well

PR #380, opened 2022-02-14
Bc2 1085 volcano set piece

IcarusPanda:
Extremely rough draft of set piece level. Will be doing some polishing tomorrow most likely.


Swiftshadow (approved):
There are no checkpoints?
Caldera edge collision is not working, you may need to place a bunch of boxes around the edge
Camera is very zoomed out, is this intententional?
Please move the script and model assets to the correct folders once final versions are complete
Level added to the select screen

PR #381, opened 2022-02-14
UI main 2022

darlynn-best:
Updated disaster select, Ben also fixed the layering issue for UI, and sandstorm assets are in. The main scripts edited are all ability selection scripts.


tcaron1024 (approved):
Fixed a few bugs
    Selecting another disaster in the same category did not deselect the first one
    AbilityManager "Show Ability Choice" checkbox works again
Made existing levels work with the new choice system
In the future, please make channels in the "Channels" folder
    Keeping AbilityCategory where it is so as to not break the existing Channels
Changed ability name text color to black for readability
Reordered levels (Andrew's request)
I think the Confirm/Help buttons should be swapped - I kept hitting help when trying to Confirm because my instinct was that the button on the right side would advance

PR #382, opened 2022-02-14
Tooltip vanishing!

BenSchu438:
The tooltip will now automatically close itself once the player has both rotated the camera (whether with keyboard or mouse controls), and after using one disaster. Also updated its font to our new standard font.

Additionally, cleaned up some old debug statements in gamemanager that I forgot to get rid of in another branch.

Affected Scripts
-TooltipBehavior.cs (new)
-GameManager.cs (removed old debug statements)


Swiftshadow (approved):
Its move the camera AND use a disaster, maybe make it either/or?
Very unobtrusive change, it works well

PR #383, opened 2022-02-15
Bc2 1282 loading screen fix and computer ID key

geekguy100:
Fixed loading screen and DB submission.
Also changed the playerpref key to something else b/c I think some are coming in blank b/c of old playerprefs that are still set on players' PCs.


Swiftshadow (approved):
It works

PR #384, opened 2022-02-15
Vfx main - Lavafall Shader

pneumacashmer:
Primary thing is the Lavafall shader which levels will need to work with me to figure out what exactly it is used for and what it needs to look like.  There is WIP of Lavaparticles but it is stand alone and _mostly done_ but still has some work to finish.


Swiftshadow (approved):
Make sure Levels knows its been merged
    I will alert them, you should as well

PR #387, opened 2022-02-15
Bc2 934 jungle ramp level

NatSmarant:
Has changes to jungle jump as well, will make a new jungle jump branch once merged for changes.


Swiftshadow (approved):
Merged main into here first to make sure things went smoothly
Thunderstorm reaching full use has revealed some places where it needs improvement
    The timing of the strikes is inconsistent, needs to be regular
        Maybe try to strike on a set delay, with obvious buildup?
    Its very possible to get the ball into a bounce-loop
        Ball gets struck down, hits the ground, and bounces back into the cloud
        Repeat the above
Only affects this level as this is the only one (currently) using Thunderstorm

PR #388, opened 2022-02-15
Sound Track & Sound Balancing

Mjsciarabba:


Swiftshadow (approved):
Finally the short BGM loop is gone :P
Sandstorm sound is pretty nice
    A "selected" sound, similar to lightnings, would be cool as well

PR #389, opened 2022-02-15
Art main

awielgo:
Mostly asset imports and asset placement in curvy bridge, jungle jump, baby's first level, and big mushroom


Swiftshadow (approved):
The new arting looks pretty good
    A lot of the blank levels feel less blank
There is still room for improvement on it, once more final geometry is set
Updated collision matrix to allow archies and any other thing on that layer to only interact with the ball

PR #390, opened 2022-02-15
Bc2 1283 vertical falling level

RobertLee-bradley:
Added a new volcano vertical based level, in unity as "Fort Squidward", but will probably have an official name like "Volcano Drop"


Swiftshadow (approved):
Not sure how I feel about the level
It was very easy
Not being able to look straight down was a major annoyance
Never needed more than wind
    Never even used geyser
Ball falls very slowly given the size of the drop
    Does not feel very twitch
Delayed for a few more iterations

Swiftshadow (changes requested):
Not sure how I feel about the level
It was very easy
Not being able to look straight down was a major annoyance
Never needed more than wind
    Never even used geyser
Ball falls very slowly given the size of the drop
    Does not feel very twitch
Delayed for a few more iterations

Swiftshadow (approved):
Much more interesting than the first iteration
End is difficult, but a fair difficult

PR #391, opened 2022-02-16
Bc2 1285 singularity stopping ball's momentum

tcaron1024:
Tried out different values for the singularity stopping/slowing the ball's momentum when it enters and ended up sticking with dividing the ball's velocity by 6 and increasing the outer pull force a bit. The amount it gets divided by is a variable that can be set in the inspector as well. It is still a bit jarring when the ball enters the singularity, but any lower and the ball can still go through the singularity if it has enough speed. 

Script:
SingularityAbility.cs


Swiftshadow (approved):
Feels a lot smoother
With enough speed (earthquake-boosted lightning wind combo) you can still barely break out
    I don't think this is an issue, as when the ball gets out it is significantly slowed
The increased outer pull makes the singularity slingshots possible again!
    Harder to aim, but that just raises the skill ceiling

PR #392, opened 2022-02-17
BC2-1075 teleporting changes camera's angle in maze

tcaron1024:
Added the ability to set the camera to a custom angle after the ball teleports using the teleport hole. Set that so the camera always angles correctly when teleporting to new islands in the maze.

Scripts:
TeleportHole.cs


Swiftshadow (approved):
Nice QOL change
    David will be happy
Camera snap can be noticeable, but its not terrible

PR #393, opened 2022-02-17
Bc2 903 lava fish

RyanSpringer1500:
Basic implantation, switched from animator base to the functions enabling the fish to jump/launched into the air.


Swiftshadow (changes requested):
You have moved the Hazards folder under the Abilities one
    Undo this please
Missing XML comments on
    LavaFish.cs line 19
    LavaFish.cs line 38
    LavaFish.cs line 40
    LavaFish.cs line 53
    LavaFish.cs line 68
LavaFish.cs line 35 should be a private
    Then missing an XML comment
Rigidbody should be set on Start()
Instead of the Run functions directly adding velocity, can they start a Coroutine that adds a force on a set timer?

Swiftshadow (approved):
Fish seems to move slowly, but that can be adjusted later - TJ

PR #394, opened 2022-02-17
Bc2 1284 bady dino functionality

Swiftshadow:
Implements basic greybox of the baby dino hazard. It detects the player, anad shoots a raycast out towards it. If the raycast hits, a force in the direction of the raycast is applied to bounce the ball away.

Modifiable vars:
    HitLayers: which layers should the raycast hit
    Shoot force: How much force to apply when launching the ball
    Time Between Shots: How long before the dino shoots again
    Hitline life: How long the LineRenderer showing the hit stays visible

Scripts edited:
    BabyDinoController.cs


tcaron1024 (approved):
Added XML comment on ShowHitline()
Fixed bug with dino firing one additional time after ball leaves radius
Code looks good
Should the dino always hit the ball if its in the radius? The raycast is pretty much instant, so I wasn't able to dodge it ever. Maybe something to talk about with designers

PR #397, opened 2022-02-17
Bc2 1087 volcano chase level

NatSmarant:


Swiftshadow (approved):
Interesting idea
Splines may need to get looked at
    Licensing issues, but they work for now
    Animation instead?

PR #398, opened 2022-02-17
Bc2 1293 consistent text spacing

QDonnelly-254:
Text on win screens and tooltip is consistent and pretty
Brown fade for pause
Changed font sizes for text on above screens and their spacing too


Swiftshadow:
Merged into ui-main-2022
I could not tell a difference in the text, will assume it was there
The brown dimming on the pause screenn looks decent
    IMO black would be better, but I am no artist

PR #401, opened 2022-02-17
Vfx main

tcaron1024:
Geyser changes to lava based on if the level is a volcano one or not - this can be set in the LevelLoader object. Also, disasters can be placed freely no matter angle.

Script changes :
GameManager.cs - added a bool
GeyserLavaCheck.cs - new script
LoadingBehavior.cs - Sets game manager based on serialized bool set in inspector


Swiftshadow (approved):
Lava geyser looks pretty dang cool
    Can the steam change color as well?
Geyser can now be a sideways launch if you try hard enough

PR #402, opened 2022-02-18
Bc2 1085 volcano set piece

tcaron1024:
Iteration of volcano setpiece level

Scripts:
Fireball.cs - New (controls fireballs shot by hippo god)
HippoGodController.cs - New (moves and shoots fireballs)
DeathCollider.cs - New (same as DeathTrigger.cs, just used for normal collision instead of triggers)


Swiftshadow (approved):
Looks good!
Checkpoints are a much needed addition
Added XML comment on HippoGodController.cs, Start() method

PR #404, opened 2022-02-19
Art main

Clkerr:
Noticed the new greyboxes were missing a few key art assets:

The correct skyboxes have now been added to all levels. Skyboxes are implemented in the loader scene for each level under the lighting inspector.

All jungle levels (water death planes) have an underground plane added specifically to hid the bottom of the islands. Helps the depth perception of the water.

All volcanic levels have been updated to have the correct island texture grid and death planes.


Swiftshadow (approved):
Lowered the underwater planes to fix visual error
Skyboxes look nice
    I honestly cant tell the difference between default and jungle though

PR #406, opened 2022-02-20
Bc2 856 database 3.0 - Statistics Display

geekguy100:
Created a script that pulls info from the tracked statistics and displays them in a text field.
I created a prefab in Prefabs/UI called "Win Panel w_ stats". Feel free to pop that in GameplayUI and modify it.
Note that it has to be linked up to actual scene transitions and stuff.


Swiftshadow (approved):
Looks good!
Added XML comments throughout

PR #407, opened 2022-02-20
BC2-1312 fixed empty game manager

geekguy100:
The loading behavior script was calling GameManager.instance during the flythrough, before it was spawned into the world. This caused a new GameObject to be instantiated with the GameManager script but all references (fields in the inspector) were missing.
I created a new script called LevelSettings.cs which holds settings unique to each level. Right now it only has IsLavaLevel. A ScriptableObject is hydrated at the start of the level that can be used by other scripts that need to access the settings.


Swiftshadow (changes requested):
Change makes time start immediatly upon loading a second or more level
    Steps to replicate:
        Launch game from title screen
        Start a level with a timer
        Use a disaster (and hit the ball with it)
        Use the pause menu to return to the main menu
        Start another level
        The timer for the second level will be going right off the bat

Swiftshadow (approved):
Bug referenced earlier is more widespread, will get a seperate bug fix branch.
Integrated the geyser color changing with the new SO setup

PR #408, opened 2022-02-21
BC2-1314 fixed timer preemptively starting

geekguy100:
Timer started preemptively when loading into any level after an initial level.
Shoutout @BenSchu438 and @tcaron1024 for collabing with me :D


Swiftshadow (approved):
Bug fixed!

PR #409, opened 2022-02-21
BC2-1315 music notes billboard towards cam

geekguy100:
The spine slide's music notes now always face the camera, preventing strange visuals.


Swiftshadow (approved):
Nice QOL change
Moved the script, added it to the correct namespace, and added some XML comments

PR #410, opened 2022-02-21
Bc2 876 connors levels

IcarusPanda:
Redid the Volcano part of the level using Eins Volcano. Redid the platforming section, added in checkpoints.


tcaron1024 (approved):
New level looks good

PR #411, opened 2022-02-21
UI main 2022

QDonnelly-254:
Text spacing fade outs and other various UI related things I can't recall right now.


Swiftshadow (changes requested):
Win screen buttons don't work
    It looks like they are not calling any functions: the button stuff is empty
Otherwise, looks very, very cool!

Swiftshadow (approved):
Stats screen now works!
Sandstorm cursor looks pretty dang cool

PR #413, opened 2022-02-22
Art main

EinEstes:
Lavafish, turtle, and hatchling prefab.


Swiftshadow (approved):
Replaced former greybox hazards with the arted versions

PR #414, opened 2022-02-22
BC2-1320 Buttons turned off

tcaron1024:
Turned off the close buttons on tutorial 1 panels. No code change


Swiftshadow (approved):
Makes changes requested by David
    X's on tutorial 1 are gone - it lied to the player
    Thunderstorms temporarily disabled - can cause a difficult-to-escape loop
        A level with them enabled is at the end of the Level Select list
On the record, I am still hesitant to do this, but fixing large issues like this is what lockdown is for

PR #415, opened 2022-02-24
Bc2 1113 falling lava rocks

brocEdson:


Swiftshadow (changes requested):
Functionality looks good, implementation a bit off
    Instead of respawning a new set of rocks each time, can you use the same one?
        Reset its position and size after the respawn time
    Can the the three scripts be merged into one?
Also missing XML comments in all extept the RockCollide.cs script

brocEdson:
Discussed in person, decided that this implementation would be sufficient after all

Swiftshadow (approved):
Discussed in-person
    Came to the conclusion there is no timely way to make the changes suggested
New terrain shapes can be added in with a new prefab and a model swap

PR #417, opened 2022-02-26
Bc2 1283 vertical falling level

RobertLee-bradley:
Adjustments were made to the Curvy Bridge Level and the Big Mushroom level based on the discussion this past Tuesday and Thursday.


Swiftshadow (approved):
Rocks are a nice addition, as are the mushrooms on them!

PR #418, opened 2022-02-27
A temp Resolution fix for the weird scaling and the full screen bug fix

darlynn-best:
The resolution issues are caused by the game not scaling from a 16:9 aspect ratio correctly. For the time being the aspect will stay in 16:9


tcaron1024 (approved):
Looks good
Resolution seems to change correctly in build now

PR #419, opened 2022-02-28
Grass texture slightly changed to have new coloration, Sandstorm arrow and visual clarity, meteor particles lessened.

SpencerStilwell:
Pog? Change is small in the grass, but I like it personally.


Swiftshadow (approved):
Personally, I'm not a huge fan of the few grass
    That said, I do like the detail it adds
Less particles on the meteor will be a good performance boost
New sandstorm arrow is nice
    Can be hard to see when pointing at the camera
    Love the falling sand particles that come off it

PR #420, opened 2022-03-01
Bc2 1319 loading screen improvements

geekguy100:
Multithreading implemented into database submission system. Along with only hazards that have been interacted with being submitted, submission is much much faster - even if all hazards were to be submitted, the process would be almost just as fast!

Since child threads cannot work in the Unity object space (i.e. they cannot manipulate Unity objects and components), the ThreadQueuer class can enqueue Actions to be executed in the main thread via Update(). This is typically being used to invoke events which manipulate Unity objects, for example the loading bar's slider.


Swiftshadow (approved):
This makes things so much faster
I did not run into issues with the multiple threads breaking stuff

PR #421, opened 2022-03-01
Gitignore updated

geekguy100:
Submitting that pull request which readds the fmod DLLs and updates the .gitignore to prevent the DLLs from getting deleted.


Swiftshadow (approved):
Hopefully FMOD will stop giving as many issues

PR #422, opened 2022-03-01
Bc2 973 big mushroom iterations

RobertLee-bradley:
made some iterations to both the big mushroom level and the dropper level.


Swiftshadow (approved):
Andrew Approved
Changes to Dropper are nice
    Lets see how they play with the increased gravity
Could not tell any changes on Big Mushroom, but there are changes in the file
    I am just blind

PR #424, opened 2022-03-01
Bc2 934 jungle ramp level

NatSmarant:
first iteration of ramp level


Swiftshadow (approved):
Andrew Approved
Flytraps were
New thunderstorm vfx
    Kept thunderstorms in due to bounce loop being mostly mitigated
    New vfx also make area much more clear

PR #425, opened 2022-03-02
Bc2 1353 physics rebalancing

tcaron1024:
Gravity doubled and disaster numbers changed to work with it


Swiftshadow (approved):
Gravity go BRRRRRRR

PR #426, opened 2022-03-03
New Storm Cloud hazard Overhaul

SpencerStilwell:
Pog? See how this works


Swiftshadow (approved):
Circle on the ground and the shadow are nice additions
If the thunderstorm is too low, the circle won't appear
    This will be made clear to the level designers
    As will the thunderstorm being more of a point-denial instead of blanket denial
Prefab is keeping the old cloud as the top of the rain column
    IMO it looks cool, and why waste the cloud?

PR #429, opened 2022-03-05
Vfx main

SpencerStilwell:
Small changes, but they were requested by art members. Lava fish is lava-ey.


Swiftshadow (approved):
Lava fish effects look pretty nice

PR #430, opened 2022-03-05
Bc2 1074 maze things

tcaron1024:
Minor adjustments to the maze - added some mushrooms to one of the islands, added Fred


Swiftshadow (approved):
Lowered the foliage
    Ashley may murder me
        Only if I don't get her first >:)

PR #431, opened 2022-03-05
Art main

EinEstes:
Color blind filter, flythrough, and more


Swiftshadow (approved):
@Ein, make sure UI/UX knows how to enable/disable the colorblind filter
Placement on hippo chase looks nice, but could still use iteration

PR #432, opened 2022-03-05
Bc2 989 volcano transition

IcarusPanda:
First pass on Environmental story telling pass charli and gwen were suggesting. Still need more animations and such, but this is a vague idea of what they were suggesting.


Swiftshadow (approved):
General functionality notes
    Swaps the jungle hippos in the water to an idle-only animation controller
    Remaps the baby dino animation to functionality
    Re-centers the baby dino sphere to make it more consistant
    Makes the volcano ankylos functional
    Removes the duplicate fmod dll

PR #433, opened 2022-03-06
Bc2 1255 three path iteration

M1sterChaos:
Greyboxes for Conveyor and GOLF are done.

Please don't delete branch, may reiterate on these levels on tuesday after feedback.


Swiftshadow (approved):
Rotated wheel to be horizontal
    It was too easy to continually miss the wheel and get stuck trying to go in

PR #434, opened 2022-03-06
I readjusted the holes

RobertLee-bradley:
on curvy bridge and big mushroom and the intro level


Swiftshadow (approved):
Hole ridges begone

PR #435, opened 2022-03-06
Art main - fixed Fort Squidward and added flythrough

awielgo:


Swiftshadow (approved):
New flythough
    Needs to pop the checkpoint
Textures in dropper look much better

PR #436, opened 2022-03-07
Bc2 1445 object pooling archies

tcaron1024:
Pteros (not archies i have the bad brain) now use object pooling - functionality should be exactly the same other than that.

Scripts:
PterodactylSpawner.cs
PterodactylHazard.cs


Swiftshadow (approved):
Looks good
    If ptero's get set to spawn faster, the pool size will need to be increased
    Dynamic pool?
        If there are objects in the pool, use them. Otherwise, spawn a new one, add it to the pool, and use
Added some missing XML comments

PR #437, opened 2022-03-08
Bc2 1306 thunderstorm bounce loop

brocEdson:


Swiftshadow (approved):
Added an indicator from one of the strike vfx components
Need to look into its behaviour on a ramp, but that can come during polish
Bolts and indocator vfx need to be pooled
    Calling our resident expert, TJ!

PR #438, opened 2022-03-08
Lightning Strike Indicator

SpencerStilwell:
Its attached to the lightning bolt prefab in a new animation it does. It's looping by default; we can change that.


Swiftshadow (approved):
RIP Thunderstorm
The VFX still look cool
    Not mapped anywhere - thunderstorm cut

PR #440, opened 2022-03-08
Bc2 1454 scorpion polish

brocEdson:
Walking animation will get in later


tcaron1024 (approved):
Scorpion looks good
Functionality works well

PR #441, opened 2022-03-09
BC2-1461 decals pool

tcaron1024:
Ability decals pool so they aren't destroyed/spawned every time you switch

Scripts:
AbilityManager.cs
IAbilityPlacement.cs


tcaron1024:
F

PR #442, opened 2022-03-09
Art main - flythroughs, changed base lizard placement

awielgo:
all preexisting flythroughs have been slightly tweaked, curvy bridge and intro should have flythroughs now, fixed disconnect between lizard model and prefab positions and shifted all already placed lizards to correct spot


Swiftshadow (approved):
New flythroughs look nice
Maze one is better, but still feels weird imo
Checkpoint on dropper does not fire, nor does the hole

PR #443, opened 2022-03-09
Bc2 1462 disaster balance/phyics tweaks

tcaron1024:
Disasters balanced with Andrew. Fixed shark bug causing ball not to die

Script:
BallController.cs


Swiftshadow (approved):
Disasters feel as powerful as they did pre-gravity change
Friction can still hamper the wind a bit, but not overly much

PR #444, opened 2022-03-09
Bc2 1255 three path iteration

M1sterChaos:
Updated geometry, raised walls for next, added more hazards to plinko, lowered all platforms so death barrier is reached quicker, added checkpoints, fixed boost pad decals, rotated and moved wheel so it's easier to hit now

Please don't delete branch just yet


Swiftshadow (changes requested):
Level not Andrew Approved
    Andrew sent details via Discord

Swiftshadow (approved):
Andrew Approved
Conveyor removed from level lineup
    Still exists in files

PR #445, opened 2022-03-10
Bc2 1367 tutorials UI scene

BenSchu438:
Significant changes to the first 3 tutorials. I believe Shelby needs to check off on this, and maybe Charli and Andrew due to it being a level/tutorial change. The changes are as follows...

Overall:
- Tutorial prompts are now much shorter and smaller, consistently sized.
- Reworded tips on all 3 tutorials.
- Optimized the code across all 3 tutorials. Including TutorialCore.cs, Tutorial1Controller.cs, Tutorial2Controller.cs. Basically took TJ's code from Tutorial1Controller and moved it to TutorialCore, so all use it now.
- Tutorial3Controller removed. This is because it has no unique functionality and TutorialCore.cs works perfectly for tutorial 3's needs.
- Tutorial prompts now have a nice little sliding animation when changing

Tutorial 1:
- Swapped ball and hole positions to reverse tutorial
- Player now starts with wind, then given meteor
- New bridge that breaks away using the falling lava rocks hazard
- Larger rock catch on the final island to make landing there easier

Tutorial 2:
- Enabled disaster select, re-enabled par timer. Note: Par timer needs to be adjusted once par times for other levels are being determined

Tutorial 3:
- Slightly reduced distance between each of the islands to compensate for the new gravity. 

Other misc changes:
- Updated the Checkpoint text to now always look at the camera and use the new font
- Fixed minor bug with LavaRocksController.cs, adding in a 'triggered' bool check


BenSchu438:
Also Tutorial 2 and 3 were swapped in level select, but instead in UI-Main-2022. This was done as Level Select is being changed there, so as to prevent any merge conflicts.

Swiftshadow (approved):
Shelby Approved
Charli Approved
Nice way of adding specific bonus functionality!

PR #446, opened 2022-03-10
dropper adjustments

RobertLee-bradley:
adjusted the dropper to be better and more forgiving in the first part, the flythrough NEEDS to be adjusted now


Swiftshadow (approved):
Andrew Approved
Re-enabled colliders on a few Maze mushrooms
Dropper flythrough does definitely need to be redone
Can the deletion of platforms on dropper be done a bit lower? At the default zoom you can see them popping out
    Assuming doing it later will still do the intended effect

PR #447, opened 2022-03-10
Added offset and rotation to lava fish

brocEdson:


Swiftshadow (approved):
Make sure the documentation has a formula/instructions to get the best jumps
    Right now, a very select few jumps look decent, there has to be a mathematical way to determine if one will work
        Document how it works currently, figuring out the math can happen later if required
Jumps look a lot nicer
    Conveyor's fish broke, but that level is being cut anyway

PR #458, opened 2022-03-24
Colorblind Mode* and fixed level selection screen bugs

geekguy100:
*Half-working colorblind mode (will continue to iterate on). Colorblind filter is stuck on the value set in the editor.

Level Selection Screen:
- Holding the left and right buttons navigates to the previous panel / next panel as intended.
- Fixed bugs with playerprefs and default values not being set appropriately.


tcaron1024 (approved):
Fixes level select bugs

PR #459, opened 2022-03-24
Bc2 1465 venus flytrap polish

brocEdson:
The flytrap should now be able to be placed above ground


tcaron1024 (approved):
Nice quality of life change
Should make placing venus flytraps a lot easier

PR #460, opened 2022-03-24
Bc2 934 jungle ramp level

NatSmarant:
booster in, par time, art pass


tcaron1024 (approved):
Looks good
Fixed an issue with one grass piece having an extra collider that stopped the ball sometimes

PR #462, opened 2022-03-24
Bc2 1369 loading UI screen

BenSchu438:
New loading screen that better matches our UI theme. Not final, but unable to get the desired art done in time so this will do for now. 

No functionality or script changes. 

Made a spinning ball unity animation.


tcaron1024 (approved):
New loading screen is much better than the old one

PR #465, opened 2022-03-25
BC2-1515 fixed order of tutorials

geekguy100:
Fixed ordering of tutorials to T1, T3, T2
Level select screen updated
Each tutorial correctly leads into the next via "Next Level" btn


tcaron1024 (approved):
kpatt got the goods

PR #466, opened 2022-03-26
Bc2 1367 tutorials UI scene

BenSchu438:
Various changes made to the tutorials. 
 - Overall, each tutorial prompt is larger, animations adjusted to accommodate new size.
 - On Tutorial 1, replaced the large rock wall to be two short rocks instead. Text updated to accommodate this change.
 - On Tutorial 1, the disasters are now loaded in as '?' and turn into the proper disaster when unlocked.
 - On Tutorial 1, fixed bug regarding manually selecting disaster through disaster panel.
 - On Tutorial 1, swapped place of wind and meteor in disaster panel to match the normal driver/putter order. Text updated to accommodate this change.
 - On Timer Tutorial, moved disaster select tip panel to the opposite side as requested. Adjusted animation to accommodate.
 
Affected Scripts:
 - Tutorial 1 Controller

There is 1 minor bug where you can see the '?' part rendered over the respawn fade which is quite obvious when going through the 'Press [Tab] to respawn' part of tutorial 1. I did fix this, but in New-UI-Main, as to prevent any potential merge conflicts with Quinn's work in the GameplayUI scene. Once that branch and this get added to main, the bug should be resolved.


tcaron1024 (approved):
I like the bigger tutorial prompts
Changes/bug fixes on tutorial 1 make it a lot better!

PR #467, opened 2022-03-26
Bc2 1275 volcano set piece iteration 1

IcarusPanda:
Added the set piece and iteration par time and also changed the starting camera on the set piece level.


tcaron1024 (approved):
Looks good

PR #469, opened 2022-03-27
Video Settings Updates

geekguy100:
Updated the video settings page in the options menu:
- Refresh rate is now taken into account.
- Resolution dropdown value is properly initialized to the screen's resolution.
- New fullscreen mode dropdown that lets users change their preferred fullscreen mode.
- Resolution options saved to a file that is loaded from on start.

Future notes:
- VSync must be disabled in order to select a refresh rate other than the monitor's max refresh rate. To accommodate for this, VSync is  disabled when the game starts. I plan to add an option in the future that lets the player choose for themselves whether vsync should be enabled or not.


tcaron1024 (approved):
All the new video options seem to work well!
Hijacked this PR to change level names and update the level pictures for level select along with some other minor bugfixes

PR #470, opened 2022-03-28
Theme Integration

Mjsciarabba:
-Jungle Theme and Volcano Theme FMOD events created
-Added appropriate theme to all levels


tcaron1024 (approved):
New music sounds good
Added BGM object to level select to fix bug where music would sometimes stop when going from a level back to level select

PR #471, opened 2022-03-28
BC2-1519 fixes bug with hippo destroying platforms after ball dies

tcaron1024:
Fixes bug with hippo in hippo chase sometimes destroying platforms after the ball dies, leaving those platforms still destroyed on the next attempt


geekguy100:
Test

geekguy100 (approved):
Phenomenal work as always, TJ! :)

PR #472, opened 2022-03-29
New wind, Cracking platform prefab started

SpencerStilwell:
Nice


Swiftshadow (approved):
Love the hippo ears
New wind vfx are cool
    IMO they can be standalone, but they still are an improvement layered
Hippo splash is very subtle, but a nice touch

PR #473, opened 2022-03-30
Vfx main

pneumacashmer:
Fixed the remainder of the hippos in the maze level to have their splash be correct.  All jungle level hippos should have their splashes aligned correctly.  Volcano still needs work, but won't get done before lockout.


Swiftshadow (approved):
Same comment about the splash as above
Used this PR to re-link the FMOD project as well

PR #475, opened 2022-03-31
Hippo Splashes Done

pneumacashmer:
With the exception of Dropper and those just in the water in the Transition Level.

Hippo God, Shark, and Ball splashes still WIP, but should be much easier to complete


tcaron1024 (dismissed):
Ball splash VFX doesn't seem to work properly - it stays on the ball even after it respawns
    It was also showing a lava splash on the big hippo jungle level

tcaron1024:
There might also be a bug with the hippo splash vfx (at least on the Dino Peninsula level). It seemed to spawn more water around the hippo's original puddle that looks a bit strange. Not sure if that's a bug or just how it looks, but something to look at

pneumacashmer:
Gave the Hippos that lost their splash their splash back.  Went through each level to make sure the death plane gave the right splash (since despite there being a water and lava death plane prefab, most levels just have the water death plane recolored)

Swiftshadow:
Ball splash VFX still stays with the ball after it respawns, leading to the splash happening on land

pneumacashmer:
All level death planes now have only 1 splash to prevent bug.

Swiftshadow (approved):
Splash no longer follow ball
It adds more than you would expect from a simple splash

PR #477, opened 2022-03-31
Bc2 1523 volcanic hippo variants

brocEdson:
Doug should check this PR out, he knows what to do with the plesiosaur


Swiftshadow (approved):
Pleiosaur collidered and set up for an environmental piece
IT IS NOT FUNCTIONAL

PR #480, opened 2022-03-31
Bc2 1466 baby dino polish

RyanSpringer1500:
finish coding launched spit for baby dino


Swiftshadow (commented):
Had to fix the spitball spawning, it was erroring out
Spitball now spawns correctly
    Rotation is a bit off, but its close enough

Swiftshadow (approved):
See above

PR #482, opened 2022-04-01
Colorblind and Graphics Quality Settings

geekguy100:
**Colorblind filter**: new setting in the Options menu under the Video tab. Select from one of many colorblind filters so graciously provided by @EinEstes! NOTE: If anything but 'Normal' is selected no Post Processing will be used. This is a limitation of the shader the filter uses.

**Graphics Quality**: Simple low, medium, and high graphics options. Using the default render pipeline assets provided by Unity. Programmers can talk through later on how these default values should be adjusted to get the smoothest experience possible on each setting. 

Settings are saved in between playthroughs. So when the player exits the game and reopens it, prev values will be restored.


Swiftshadow (approved):
Merging main in broke some of the build settings
    Fixed that
Medium and Low graphics settings make singularity and earthquake almost unusable
    Singularity does not show its radius very well, creates giant grey sphere upon use
    Earthquake does not show direction line
Kudos to yall and Ein on the colorblind filters!
Changed some spellings on the credits screen
    Skylar to Skyler
    Frank to Frak
Added people to Special Thanks
    Annabelle Pries, Matthew Kirchoff, Marissa Torrison
Gallery added
New logo and splash screen added and implemented

PR #484, opened 2022-04-01
Bc2 1522 animatics gallery

tcaron1024:
Gallery scene added with videos and sound - click on the center panel to play the videos 

Scripts: GalleryController.cs


Swiftshadow:
Merged to ui-main-new
Added buttons to unlock and reset the level progression
Maybe this scene can be repurposed into an asset showcase?

PR #487, opened 2022-04-02
Bc2 1413 subtitle system

BenSchu438:
Firstly, fuck FMOD.

Subtitle and dialogue system are in. We currently only have dialogue in for the first 3 tutorials. Here are the affected systems:

Scenes:
 - Gameplay UI. Did multiple bug fixes regarding screen render order as well as prepared subtitle/subtitle controller
 - T1,2, and 3 UI. Added and organized subtitle systems
 - OptionsPausedMenu. Added dialogue slider as well as an option to mute subtitles (which doesnt work on tutorials as we wanted those forced. Disclaimer added under setting)

Scripts:
 - SubtitleData. New SO that contains the file for the audio clip as well as the subtitle equilivant
 - Subtitle. Parent class for the subtitle trigger system. Have a plan to make the group system more optimized after this milestone
 - Subtitle Trigger/Subtitle Collider/Subtitle Call. All similar in that they inherit Subtitle and just have different ways for it to be called
 - SubtitlePlayer. Has a queue of voice lines that get passed into it. This player then plays audio and subtitles for each item in the queue in order. 
 - SoundSettings (options). Adjusted to handle the new audio settings. 
 - T1 Manager. Fixed bug where pausing and unpausing can unlock disasters before tutorial unlocks them

Known Bugs:
 - If the player goes into options menu and changes slider for the dialogue slider while a dialogue line is being played in level, then upon unpausing and returning to the level, that voice line will not continue. Fix planned but requires more work and I dont wanna break FMOD again.

Finally, fuck FMOD.


Swiftshadow (approved):
FMOD be dumb
Nice work on the subtitiles and voice overs!!!!!!

PR #488, opened 2022-04-02
Art main - level select pics, tutorial flythroughs, slight environment tweaks

awielgo:


Swiftshadow (approved):
New screenshots are very nice
As are tutorial flythroughs
    And the changes to Jump and Windmill's
Disabled dropper's lavafalls
    Could not solve their lightning issue, will look into during polish

PR #489, opened 2022-04-03
BC2-1546 stops fmod event when going back to menu

tcaron1024:
Fixes gallery audio continuing after going back to main menu

Script:
GalleryController.cs


Swiftshadow (approved):
Works well!

PR #490, opened 2022-04-04
Camera move

M1sterChaos:
Moved camera for windmill

Please don't delete branch or I will be very sad


Swiftshadow (approved):
New level orientation is set

PR #491, opened 2022-04-04
BC2-1547 shark ball hole eat fix

tcaron1024:
Shark no longer eats ball in hole.

BallController.cs


Swiftshadow (approved):
Shark no longer eats the ball in a hole
Rotated the camera on Margaret's Meal to make it face the hole
Moved some trees in Pyramid Scheme

PR #492, opened 2022-04-04
BC2-1548 renamed wrongly named hazards, added trackers to newer hazards

geekguy100:
Renamed some hazards named "magicmanjohnson" and one other similar name to "Hippo" (since they were both Hippos).
Added tracking to Shark
Removed some delimiters in names that were placed too early.


Swiftshadow (approved):
Better tracking will make using the information easier!
Idea: would it be possible to have the level tool change the name as it spawns to the correctly formatted one?

PR #493, opened 2022-04-05
Removing Dropper Squares and Updating Venus Fly Traps in My Levels

RobertLee-bradley:
I removed the useless lava squares in the dropper level as well as adjusted the vertical placement of Venus Fly Traps based on changes made by Broc and the other programmers.


Swiftshadow (approved):
Flytraps look a bit weird IMO, but thats just my opinion
Otherwise looks good!

PR #494, opened 2022-04-05
margret fixed

NatSmarant:


Swiftshadow (approved):
Margret no longer clips the camera!

PR #496, opened 2022-04-06
Bc2 1549 animation offsets for creatures

brocEdson:


Swiftshadow (changes requested):
I did not know you had a script for offsetting animations already made!
Changes themselves look good - they did not apply to all instances of the objects
    Dropper level the hippos do not use the prefab, they will need to be manually replaced
    Check all the volcano hippos, they all probably need to be replaced

Swiftshadow (approved):

PR #498, opened 2022-04-06
Flytrap placements updated

NatSmarant:


Swiftshadow (approved):
Flytraps no longer clip into the ground!

PR #499, opened 2022-04-06
BC2-1593 fix maze flytrap placement

tcaron1024:
No code changes, fix maze flytrap placement


Swiftshadow (approved):
As with the above, flytraps no longer clip into the ground!

PR #500, opened 2022-04-07
Made idle triceratops

brocEdson:


Swiftshadow (approved):
Envrionment hippo!!!!!!!!!!!!!!!!!!!!
    Added offset script to make sure they are not robotically in-sync

PR #501, opened 2022-04-07
Vfx main

pneumacashmer:
DeathTrigger.cs had to be changed for the sharks to ensure they spawn the splashes at the right spot regardless of where the death plane was placed.  SharkManager.cs now has a check to find the right GameObject which is the deathplane


Swiftshadow (changes requested):
Interesting approach to finding the deathplane
    I'm unsure of the performance vs finding a tag
        There are fewer deathtriggers than objects, but it needs to loop over the scene to find them anyway
        Do some experiments on this, and use the more performant option
            If the options are negligable, then use tags as other areas may need to reference the tag
I don't see the splash when the shark comes up (Intro level)

pneumacashmer:
Shark Prefab fixed as there was an issue with the original find.  Sharks in lava levels were the wrong prefab so changes in there was simply swapping them out.

Swiftshadow (changes requested):
Lava gets a color-altered buzzsaw as per Art team
    Fixed and pushed
    Great White is only the 3 beach levels
Not seeing the splashes on the great white (beach level only)
    Should be water splashes, great white is only beach level
Splashes on Volcano Setpiece are not spawning at the correct height
    The entire level is on an incline

pneumacashmer:
Shark now has a rigidbody as to handle when it collides with the deathplane.  That is used to better time out when to spawn the effect even with sloped death planes.  With the sharks being corrected to the right level, they had the wrong splash and scene changes are fixing them.  The shark hazard in Volcano Setpiece Level was also lowered to below the deathplane.

Swiftshadow (approved):
Applied a tag to the deathplanes and used that for finding them
Splashes work, and look pretty nice!
New jungle water also looks pretty cool
    Dropshadow appears a lot more prominately on it

PR #502, opened 2022-04-07
Art main - Flythrough tweaks, hippo replacements, pillars improved

awielgo:


Swiftshadow (approved):
Pillers look a bit weird IMO
Same with hippo head statues
Re-applied the animation offset script, it got snapped somewhere
Will take your word on the de-janking of flythroughs
    They work, I cannot see a difference

PR #503, opened 2022-04-07
Bc2 1598 breakable terrain colliders and functionality

brocEdson:


Swiftshadow (approved):
Thats a lot of rocks
Optimization for the future, the more regularly shaped pieces can get simpler colliders
Also the curved islands have a heck of a lot, they may need to be reshattered but smaller
    Dependant on how many actually get placed/used

PR #504, opened 2022-04-08
Bc2 1259

M1sterChaos:
Made the changes to windmill for cheesing the sides of the level.

For GOLF I had TJ make the earthquake not permeate when teleporting, added the pin counter, and changed the walls of the nest to be better.

Ashley is changing the hippos on plinko for me, they are already in progress as you read this

Please don't delete branch yet.


Swiftshadow (approved):
Its possible, but unlikely, to skip out of the hole after earthquake gets cancelled but before the teleport
Idea on GOLF (specifically to help speedrunners)
    If the player enters a hole within X seconds, their chance to get teleported to the bonus windmill level is 0%
        The random bonus level will annoy speedrunners, and make TAS's nearly impossible

PR #505, opened 2022-04-08
Credits + Options screen polishing

geekguy100:
Credits Screen:
- Reordered some names
- added in a thank you msg at the bottom
- increased scroll speed

Video screen:
- re-positioned UI elements
- changed drop-down text font
- changed drop-down text color
- increased scroll speed on drop-downs
- increased length of drop-down container
- slightly increased title size
- shrunk wooden panel so it doesn't take up too much screen space

Audio and Mouse screen:
- shrank title font size
- added in vertical layout groups for more precise and consistent organization

All screens:
- Adjusted anchors for better 16:10 support

It looks like @darlynn-best also had some WIP additions. But I am unaware of exactly what those entail.


Swiftshadow (approved):
To look into:
    OnMouseMove() (and the InputAction its on) may be the reason moving the mouse moving is causing a performance hit
New font on the options dropdowns is nice
Re-ordered credits are also nice
    Rainbow "Thanks for playing!" is a bit hard to read - yellow on beige
        Looks out of place
Subtitiles need to be looked at
    Still block the cursor despite being on the IgnoreRaycast layer

PR #506, opened 2022-04-09
BC2-1594 Camera changes done with ui

tcaron1024:
Made camera movement get adjusted by options menu, added different sensitivity sliders and inversion toggles for keyboard/mouse movement. I need feedback on if the current inversion settings work as would be expected. They are how I would expect but not sure if that's correct

Scripts:
CameraControls.cs - Kept the base of the camera controls, mostly changes in RotateCamera()
CameraOptions.cs - added more sliders/toggles, made the settings save with PlayerPrefs


Swiftshadow:
Merged to ui-main-new
New options are nice to  have
My two cents on the invert (still get more input, specifically Shelby):
    Mouse works as expected
    Keyboard Y is inverted with the check turned off

PR #507, opened 2022-04-11
Bc2 1552 hippo god speedup

geekguy100:
A speed multiplier added to the Hippo God in the chase level has been added.

How it works:
In the Inspector on the LavaChaser are two new fields:
- Distance Speed Multiplier: a multiplier added to the base speed. This value is **multiplied with the difference** in percent along the path of the player and the hippo god. For example, if the Hippo God was at the very start of the path and the player in the middle, the difference would be 0.5. If the player was at the very end of the path and the Hippo God at the start, the difference would be 1.
So to clarify...
    - 1 represents the end of the path, and 0 represents the beginning.
    - This is the main value that should be adjusted by designers.
- Speed Adjustment Time: This a field that is there in hopes of optimizing performance. The method I'm using to determine the golf ball's point along the spline is costly. This value can be changed to adjust the frequency at which speed modifications are made to the Hippo God. Obviously, lower values will be more realistic and create smoother speed ups / slow downs, but performance will suffer.

Right now, the Distance Multiplier is at 100. This is an arbitrary value I chose. After some brief testing, I found it works decently well but should be tested further.


Swiftshadow (approved):
Looks good
    Performance notes are nice to know!
        If it becomes a problem, people can look into alternative ways to achieve the same goal
We may not be keeping the spline tool, in which case this method will need to be refactored
    Won't be until after graduation at the earliest

PR #508, opened 2022-04-11
BC2-1607 Makes the shark no longer doublefire when over the water for…

Swiftshadow:
Prevents the shark from restarting its process midair if the player manages to evade it long enough.

Scripts modified:
    SharkController.cs
        Added a null check for the jump coroutine when starting the jump, and nulled it out when ending the jump


tcaron1024 (approved):
Nice bug fix

PR #509, opened 2022-04-11
BC2-1620 Makes the volcano fireball prefab functional

Swiftshadow:
Place it in the world, make sure all the VFX are lined up to appear, and it will work!

Scripts Created:
    LavaFireballController.cs
        Used to enable/disable the death trigger, and speed up/slow down the animation if desired


tcaron1024 (approved):
Volcano explosion looks great
Speed controls are nice as well

PR #511, opened 2022-04-12
Bc2 1563 gameplay UI

QDonnelly-254:


Swiftshadow (approved):
Merged into ui-main-new
Locked selection pop-up is nice
Lava scorpion appears as "Lava" on the Arch Enemy section
Flipped end screen buggons

PR #512, opened 2022-04-12
Bc2 1610 mushroom iterations

RobertLee-bradley:
Made the mushroom level better


Swiftshadow (approved):
Rock blocker on the right side makes the level a lot more interesting

PR #513, opened 2022-04-12
Water Dropshadow Fix

SpencerStilwell:
Sort of a bandaid fix, It's the best I could come up with. Now there are water ripples on the surface of the water where the ball flies over it, making an indicator equivalent to the drop shadow only over the water.


Swiftshadow (approved):
Water ripples can form a weird line when dying in the water
    Like with the ball trail, probably disable when teleporting the ball
Added a destroy after delay to the feather particles
    They never destroyed otherwise

PR #515, opened 2022-04-13
Readded Lava Splash to Volcano Levels

pneumacashmer:


Swiftshadow (approved):
Splashes are back on the lava!

PR #516, opened 2022-04-13
UI main new

darlynn-best:
Lots of changes to gameplay UI win screen, subtitle fixes, level memory, and level select tutorials.


Swiftshadow (changes requested):
Win screen changes are nice
Mouse settings are also nice
The T3 level select tutorial should only happen once
    If the player finishes T3 and hits "Next Level" after seeing the tutorial once it should just roll to the next level
Cooldowns don't show up on the disaster selection bar

Swiftshadow (approved):
Love the changes made!
Tutorial glow is nice, but after seeing it once it should be disabled
    Run through Shelby first though
See above comments marked "Merged into ui-main-new" for comments on this

PR #517, opened 2022-04-14
Bc2 1622 disaster cooldown reduction changes

tcaron1024:
Tweaks to disaster balance:
- Meteor size increased and power slightly increased
- Earthquake duration/cooldown lowered, power slightly increased
- Lightning cooldown increased, cooldown reduction when using lightning decreased
- Cooldown reduction when using geyser decreased

Made the amount that specific disasters decrease the cooldowns of other disasters changeable per disaster

Scripts:
AbilityData.cs - add cooldown reduction variable
CooldownManager.cs - Make cooldown manager get the list of active disasters at game start then use that list to decrease cooldowns appropriately


Swiftshadow (approved):
Lightning is no longer the definitive best disaster to use
    Its still strong, but other combos can get across certain levels faster/more easily
Change to the cooldown reduction amounts will give another lever to pull when balancing!

PR #518, opened 2022-04-14
Bc2 1522 animatics gallery rework / polishing

geekguy100:
The Gallery is now super spicy with new features.

- The first frame of each video is displayed instead of a blank screen or the last frame viewed.
- Play icon is displayed on the screen when the video player is not playing anything. (thx Scottie 😍)
- Video is tinted gray when it is not playing anything.
- Video can be paused!
    - If the video is playing and is left clicked, the video will pause. If the video is paused and it's left clicked, it'll play!
    - Audio is paused as well!
- When the video reaches its end, the video player is reset. i.e. the first frame of the vid is displayed, and the video player is paused with the appropriate paused UI on-screen.


Swiftshadow (approved):
A pausable video player? *Gasp*
Nice implementation, makes it a lot clearer what the gallery is for

PR #519, opened 2022-04-14
Bc2 1599 shark polish

brocEdson:
Shark returns to water and has a telegraph mode where he sticks his head out of the water before attacking


Swiftshadow (changes requested):
Shark appears to be targeting above the ball, not the ball itself
    Makes it much easier for the shark to miss, and it just looks weird for the shark to be floating upwards
What is the telegraphing? I was unable to see it

Swiftshadow (approved):
Shark does not look as jank anymore!
Telegraphing is nice
    Potential issue with the shark missing if the player is flying really close to the water, but its not common enough to warrent attention right now

PR #520, opened 2022-04-14
BC2-1608 Prevents spamming spacebar on the flythroughs from loading m…

Swiftshadow:
Flythroughs now have a boolean lockout when being skipped to prevent multiple loads occurring.

Scripts modified:
FlythroughController.cs


tcaron1024 (approved):
Looks good!
Small change, but nice bug fix

PR #521, opened 2022-04-14
Bc2 1259

M1sterChaos:
Finalized the GOLF level, fixed terrain issue on windmill

Branch can finally be safely dismantled after all this time


Swiftshadow (approved):
I can't see what is different but I'll trust it is
Nothing appears to be outright broken

PR #522, opened 2022-04-14
Made the lizard do the funni again

brocEdson:
AKA fixed the lizard not colliding with the mushroom and added extra effects for it


tcaron1024 (approved):
On a lot of levels, the hippos suck the lizard in and they get stuck inside the hippos mouth
    Not necessarily a bug, but not sure if thats something we want happening since the player doesn't see the lizards as much
Messing with lizards is fun
Made lizards in the GOLF bowling section get moved by disasters too, added some small code changes to how the number of pins knocked is detected
    Increased the mass of these lizards to 1 so that they don't go too fast to go through the pin knock check colliders

PR #524, opened 2022-04-14
Bc2 1621 buzz shark bug

geekguy100:
Forgot to open this pull request, but I fixed the stack overflow bug when player dies from shark.
The EventInteractable component on the Shark was calling itself, causing an infinite loop. The correct reference has been assigned to the Event.


Swiftshadow (approved):
Shark should no longer crash stuff, yay!

PR #525, opened 2022-04-14
Bc2 1560 tutorial

BenSchu438:
Various tweaks to the tutorials, namely bug fixing and new animations for panels. New animations are designed to resolve issues of people not noticing them, so now they come down around the center before nesting in the side, as well as an occasional shake of a panel every few seconds. Renamed the tutorial levels to properly represent their order (combo tutorial is actually named T2 now!).

Modified Scripts: 
-TutorialCore.cs ~ Edited to manage the new animations 
-TooltipBehavior.cs ~ Edited so it looks for right-mouse input instead of middle mouse for mouse movement 

Future changes to the tutorial levels (namely 2, the combo one) are still planned, so please don't snap this branch.


Swiftshadow (changes requested):
Merged into ui-main-new
Holding right click while beating a level locks the cursor to the center of the screen
    Also locks it as invisible
    This softlocks the player - they cannot move their cursor, or interact with any GUI
Love the tutorial animations!
New title screen pictures are also nice

Swiftshadow (approved):
Merged into ui-main-new
Looks good!

PR #526, opened 2022-04-15
BC2-1668 Turn off normal drop shadow when over water

tcaron1024:
Turns off the normal drop shadow when the ball is over water so that it doesn't give incorrect information to the player.

Currently doesn't work on Big Hippo, Ramp, and Windmill jungle levels because they have the wrong water plane set up (doesn't work on main either). Didn't want to make those changes here because of potential conflicts, so I contacted Ashley about it 

Scripts:
BallDropShadow.cs - made the water drop shadow move with the ball separately (was previously parented to normal drop shadow) and added a channel reference to turn the normal and water drop shadows on/off when ball is over water/land
BallController.cs - raises a channel's event when the ball first goes over water/land


Swiftshadow (approved):
Works really well!
Removal of dropshadow makes the water height much easier to tell

PR #527, opened 2022-04-15
Bc2 1669 shark movement rework

Swiftshadow:
Swaps the shark from using speed-based movement to time-based movement. This means the shark will reach the player in a set time, and never miss. Removes the death trigger, replaces it with a straight call to the death channel. Makes the shark telegraph with its fin instead of mouth.

Scripts modified:
SharkController.cs


tcaron1024 (approved):
Sharks be working - need some polish but good for now

PR #528, opened 2022-04-15
UI main new - Updated Gallery Play Button 

geekguy100:
Updated the Gallery's play button with re-colored version that matches the game's color scheme.
Before it was all black, now it is a browner / tan color.


Swiftshadow (approved):
Button fits the art scheme!

PR #530, opened 2022-04-17
Updated Black Hole Visuals

SpencerStilwell:
New Space Particles, trippy


Swiftshadow (changes requested):
Falling Lava Rocks throw a Null Reference and break when gonig over them
    "Effects to Spawn" is not set
The singularity effect does look nice, but the above does break T1 completly

Swiftshadow (approved):
Dummy object used as per Discord discussion

PR #531, opened 2022-04-18
Bc2 1500 hippo chase bar

RyanSpringer1500:
review code for improvements. Hippo chase tracking code. 3 scripts created.
Waypoint, BallCheckpoint, HippoChaseBar.
names might change after review of code.


Swiftshadow (changes requested):
Unable to test, there was no test scene or area to check functionality
Missing some XML comments on functions
Incorrect format for class header comments - should be XML comments

Swiftshadow (approved):
Functionality looks good
I like the drop down animation

PR #532, opened 2022-04-19
Pyramid scheme iterations bc2 164

IcarusPanda:
First big iteration complete. Need to talk to TJ about smoving this hippo a little bit.


Swiftshadow (changes requested):
Items below marked with * are the ones that got it rejected

Level is extremely difficult without any aerial disasters
    I had to abuse physics, ramps, and lightning to complete it
Grass on moving hippo's backs looks weird
    The patches are worldspace, so the grass moves through them instead of them moving with the grass
Singularity does not move with hipos
    Don't know if it should either, thats definitely a question that needs to be discussed
*Please move all scripts, animations, and other non-scene files to their respective folders. Scenes should be for scenes only
Hippo_Parade_Buffer.cs follows none of the standards
*Final half, the noncircular platform: if you hit the slant, you get punted through the deathplane without dying
*Crystals cannot have disasters placed on them
    Again, don't know if this is intentional or not, but it is currently the case
*Splash VFX got unset, talk to Pneuma on what object to assign
Tripple fireball is cool!

Swiftshadow (approved):
Issues addressed!

PR #533, opened 2022-04-19
BC2-1543 now using GameManager timer for win screen

geekguy100:
Using the GameManager's elapsed time instead of the statistics displays for similarlity.


Swiftshadow (approved):
Time desync removed!
Also centralizes time tracking, which is a good thing

PR #534, opened 2022-04-19
Bc2 1530 anklyo adventure polish

RobertLee-bradley:
Art Pass Pull Request (From Nat, since her computer be not working)


Swiftshadow (approved):
Level looks a lot better with an art pass!
Ankylo's are a bit more consistent
    Work still  needs to be done, but as a low priority

PR #536, opened 2022-04-20
Bc2 1693 shark adjustments

tcaron1024:
Shark does the good things instead of the bad things it used to do

Script:
SharkController.cs


Swiftshadow (approved):
Sharks moves and telegraphs better!

PR #538, opened 2022-04-22
Bc2 1624 hippo god spline rework

geekguy100:
Incorporated new spline system from CatlikeCoding and got the Hippo God chase lvl working w/ it.
Wrote some documentation to guide other devs with using the new system, getting percents, etc.

- BezierSpline component must be added to game objects to create a spline.
- SplineHelper script is a helper class that contains two methods to aid in the approximation of percentages along a spline. More details in the documentation.

Please note that i did _not_ modify the Lava Chase Geometry scene. Therefore, if you go and test this there will be two Hippo Gods - one in the geo scene and the other in the hazards. After other ppl get their changes in, we'll need to delete the Hippo and old spline from the Geo scene and keep the stuff in the Hazard scene.


Swiftshadow (approved):
Nice job with the spline stuff!
    That looks like a lot of fancy math, and it works!
The double hippo does need to be addressed, it will be once Broc's geometry changes make it through
    In the mean time, I disabled the geometry hippo. Should merge conflicts arise, these changes can be lost
        This is to ensure the level can be played should the geometry changes take longer than anticipated
Dreamteck can be pulled once the geometry changes are through

PR #539, opened 2022-04-23
BC2-1688 Disables the splash on the volcano transition raised deathplane

Swiftshadow:
Also swaps the save system from concatenating strings for its path to using Path.Combine() to avoid potential cross-platform issues.

Scripts Changed:
SaveManager.cs - Swapped the string concatenation for paths with Path.Combine()


tcaron1024 (approved):
Good catch on the transition death plane spawning waves
Save changes work and hopefully will stop any cross platform problems

PR #540, opened 2022-04-23
GoldMaster1

EinEstes:
Shader work and flythroughs


Swiftshadow (approved):
Shader precision change will help with performance
    It is not the be-all-end-all, but every little bit helps
New trees look very nice!
    They do fit in a lot better
Only noticed a change to the maze flythrough, but they all looked good!
Fixed an issue where static greens on the hippo parade did not move, or collide

PR #542, opened 2022-04-25
Bc2 1697 setpiece iteration

Clkerr:
Magaret's Meal now has all the rocks fixed so it should be completely done. 

Pyramid Scheme has undergone a major art update with the fixed temple and updated environment assets with a crystal wall to help split up the two paths.


Swiftshadow (approved):
New rocks in Margaret's Meal are nice
    Very subtle change, but one thats well worth it 
    Can no longer be succ'd over the rocks
Volcano Setpiece V3 is very cool
    New temple looks a lot more spacious than the old
    Both routes are interseting and offer cool challenges
        I still prefer to skip them straight down the middle
    A new flythrough will be required, that can come on art-main

PR #543, opened 2022-04-25
Bc2 1560 tutorial

BenSchu438:
Adjustments to Tutorials, namely, a revamped Tutorial 2!

Changes:
 - Tutorial 2 revamped. Now starts the player with geyser and wind only, and then a new slowmo effect when player is trying to cross the gap with meteor and lightning to make that step of the tutorial easier 
 - Tutorial 1, checkpoint sign turned around to face player's starting position. Additionally, with the new sand effects, I make the blocks immediately de spawn as the new effects do well to hide that transition. 
 - All tutorial panels now auto vanish when completing the hole
 - Fixed bugs relating to tutorial stuff introduced with new pause panel animation
 - Fixed bug where pausing and unpausing would enable both video player and disaster 'aiming' on the win screen
 - Fixed water render bug where it would fight with other transparent VFX
 - Fixed issue where hole would occasionally render conflict with the grass by slightly increasing mask height 

Affected Scripts:
 - SlowmoField.cs ~ New
 - TutorialCore.cs ~ Checks when in win screen before enabling next panel
 - Tutorial1Manager.cs ~ Fixed input-enabling bug
 - Tutorial2Manager.cs ~ Fixed input-enabling bug
 - AbilityManager.cs ~ Added check to see if winscreen active before enabling disasters after pause
 - AbilityVideoPlayer.cs ~ Added check to see if winscreen active before enabling disasters after pause
 - GameplayUIBehavior.cs ~ Fixed input-enabling bug in tutorials


tcaron1024 (approved):
The tutorial 2 changes are great
   It will need another art pass/flythrough/picture for level select
Slow mo effect is really well done
Do we want the empty par medals showing up on the win screens of tutorials 1/2 if there is no par time?
   Not sure, something to consider

PR #547, opened 2022-04-26
Hippo Bois

Mjsciarabba:
HG in setpiece and chase make sound :)


Swiftshadow (approved):
HIppo God SFX are there!
    On Hippo Chase they seem a bit repetetive, but other than that its good!
    Also they don't sync with the flythrough animation on that level

PR #548, opened 2022-04-26
Art main - foliage updates and setpiece tweaks

awielgo:


Swiftshadow (approved):
There are still floating rocks in the setpiece
Other changes look nice

PR #549, opened 2022-04-26
Darker Drop Shadow. New Splashes for water. Fixed water jittering bug.

SpencerStilwell:
pog pog pog.


Swiftshadow (approved):
Lava splash added for lava levels!
Dropshadow now more visible over lava

PR #551, opened 2022-04-28
Art main

Clkerr:
All art passes have been done we should be all set!


Swiftshadow (approved):
All the part passes look nice
    I cannot tell the difference on some, but im sure its there
Lavafalls finally back in!!!
Temple texture looks really nice
Tweaks to GOLF's teleholes will make speedruns much nicer
Pyramid Scheme still needs a new flythrough
Skipping the final cutscene does pop up the win screen, but cutscene still plays in the background
    It should jump to the end/final frame of it

PR #552, opened 2022-05-01
BC2-1723 fix ability cooldown bug

tcaron1024:
Fixes bug where abilities may sometimes have their cooldowns not get reset properly. Bug happened if 2 (or more probably) abilities had their cooldowns end at the exact same FixedUpdate frame. The way the channels were being removed from the list did not properly work if more than 1 needed to be removed at a time. 

Script:
CooldownManager.cs (changed for loop that removes ability's cooldowns in TrackCooldowns())


Swiftshadow (approved):
Glad this was located and fixed!
The implementation looks solid, I am not precise enough to accurately test it

PR #554, opened 2022-05-03
Bc2 1368 convex colliders

geekguy100:
Added convex colliders to (mostly) the round islands in each scene. Hopefully this will contribute to performance improves as the ball / hazards collide with these updated colliders!

I noticed the prefabs themselves didn't have any collliders on them, but the prefab instances (i.e. the game objects in the scenes) did. I could've updated the prefab, but I didn't wanna mess anything else up, and then each prefab instance would've had two colliders and I woulda had to delete one and update the other. So I just went through each scene and made each collider convex where applicable.


Swiftshadow (approved):
Performance is helped, if slightly

PR #555, opened 2022-05-03
BC2-1720 fixed medium and low graphics settings causing screen issues

geekguy100:
I found out the problem was with the UI camera not taking depth and opaque textures into account. UI cam prefabs have been updated. This solved the problem.
I also turned on Depth + Opaque textures for medium and low settings b/c without them the black hole and some other disasters were invisible


Swiftshadow (approved):
The game is playable on lower graphics settings!

PR #557, opened 2022-05-03
Bc2 1725 final cutscene bug fixes

tcaron1024:
@geekguy100 did some stuff to fix the final cutscene bug 

Made the fade for the intro cutscene longer - had to move the scene loading for the intro cutscene out of FinalCutsceneHandler into a new script

Scripts:
FinalCutsceneHandler.cs (added bool to disable scene loading on video end)
LoadSceneAnimationTrigger.cs (new script, probably could have reused another one but this was faster)


Swiftshadow (approved):
Final cutscene bug has been fixed!
Fade to intro level is definitely longer

PR #558, opened 2022-05-03
Bc2 1710 ball physics material bug

brocEdson:


Swiftshadow (approved):
It works, the ball's material is set on every scene load!
    May be set a bit too often (lots of scenes get loaded at once), but they are never loaded during gameplay

PR #559, opened 2022-05-03
Lava floating particles updated. Earthquake visual decal less sharp o…

SpencerStilwell:
…n the end.

Lava floating particles completely overhauled in how they look and spawn. Earthquake update was what David mentioned in class today.


Swiftshadow (approved):
Lava particles no longer spawn 900 individual objects

PR #560, opened 2022-05-04
Sound main

Mjsciarabba:
lots of new sounds implemented


Swiftshadow (approved):
The new sounds are subtle, but very nice!
HIPPO GOD ROAR IS NOT CONSTANT! ITS NO LONGER ANNOYING!

PR #561, opened 2022-05-04
Fixed layering bug causing the scorpion not to interact with the ball

brocEdson:


Swiftshadow (approved):
Scorpion be workin again!

PR #562, opened 2022-05-04
Fixed hippo god sound carrying over on final scene

geekguy100:
Removed event from HippoController, added in Matt's TriggerSound script, and added the event in there. Now the hippo god sounds don't play when the final cutscene loads!


Swiftshadow (approved):
Nice catch, nice fix!

PR #565, opened 2022-05-05
Bc2 1360 general optimizations

Swiftshadow:
Bakes lightmaps and occlusion maps for all the levels. This should help performance a good bit.
Fixes a bug with the always unlocked levels where they were not always unlocked
Fixes a bug with trhe dropshadow appearing below the map. Moved it up so 1/8 is above the ball, the rest is below

Scripts modified:
LevelData.cs
BallDropShadow.cs


Swiftshadow:
Looks good
    Lightmaps cut
- TJ

PR #566, opened 2022-05-05
Bc2 1755 singularity sfx bug

BenSchu438:
Fixed a few bugs.
- Fixed bug where ball hit sound would spam on singularity. Did this by disabling ball hit sounds while in singularity and enabling again once ball leaves singularity or singularity ends.
- Fixed bug where the trail renderer would draw a line when respawning 
- Fixed bug where the ball fire visuals and ball lightning visuals were still active on the ball after respawning
- Fixed bug where the ball would occasionally give off a dark blue shadow

Affected Scripts:
- SingularityAbility.cs
- BallController.cs
- PlayerRespawnManager.cs


tcaron1024 (approved):
Nice QOL fixes

PR #567, opened 2022-05-05
Bc2 1757 ball shadow and other bugs

tcaron1024:
Sized up ball in t2 so it matches other balls, added in scene offset to drop shadows in t1/t2 so they appear properly. Ben made the teleport holes turn off the ball's trail so it doesn't look weird when teleporting.

Script:
TeleportHole.cs


Swiftshadow (approved):
Minor tweaks, but worthwhile one!

